PR #2, opened 2014-07-28
IMAGING-138  Provide some Ant tasks to perform image conversions

ghost:
Adds a single Ant task of image conversion, tested on a small PNG file converted into ICNS.


garydgregory:
If we want to support Ant, we should split Commons Imaging into a Maven multi-module project and put this into a new module to create jar called `commons-imaging-ant` IMO.

PR #4, opened 2015-01-15
IPTCParser now takes the raw bytes of an IPTCRecord element into account

ola-github:
The IPTCRecord supports a String value and raw bytes. When writing an IPTCRecord into the file the raw bytes are ignored and the string value is used only instead. This value is encoded in charset 8859-1. There is currently now way to use a different encoding then 8859-1.

The given change now uses the raw bytes if existent. if not then if falls back to the previous strategy (string value encoded as 8859-1)

My context:I am currently working on handling of IIM and XMP metadata in images and i need the capability to use a encoding different from 8859 im IIM. I would be happy to hear your opinions on the proposal.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/jpeg/iptc/IptcParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/jpeg/iptc/IptcParser.java
@@ -438,10 +438,20 @@ public class IptcParser extends BinaryFileParser {
                 }
                 bos.write(element.iptcType.getType());
 
-                final byte[] recordData = element.value.getBytes("ISO-8859-1");
-                if (!new String(recordData, "ISO-8859-1").equals(element.value)) {
-                    throw new ImageWriteException(
-                            "Invalid record value, not ISO-8859-1");
+                /**
+                 * favor raw bytes over value. This allows callers to use their
+                 * own encoding of fields.
+                 */
+                final byte[] recordData;
+                if( element.getRawBytes() != null && element.getRawBytes().length > 0 ) {
+                    recordData = element.value.getBytes();
+                }
+                else {
+                    recordData = element.value.getBytes("ISO-8859-1");
+                    if (!new String(recordData, "ISO-8859-1").equals(element.value)) {
+                        throw new ImageWriteException(
+                                "Invalid record value, not ISO-8859-1");
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


mgmechanics:
Hi Oliver,
Thank you for your patch! 
I'm collecting all the waiting patches for Apache Commons-Imaging and employ them at http://sourceforge.net/p/albonubes/
I hope you forgive me that I use your patches for a child project. Maybe you give it a try?
Michael

kinow:
I think from your description you have a valid use case for handling the encoding yourself, instead of letting the library use ISO-8859-1 only.

I started to rebase the branch locally, when I realized that the `IptcRecord` class actually lost the `#getRawBytes()` method. So we can't really merge the code any longer.

Not sure why it changed, but looks like we will have to either add it back, or find an alternative solution. Sorry @ola-github Let me know if you have any idea how to update the pull request.

src/main/java/org/apache/commons/imaging/formats/jpeg/iptc/IptcParser.java
```diff
@@ -438,10 +438,20 @@ public int compare(final IptcRecord e1, final IptcRecord e2) {
...
-                final byte[] recordData = element.value.getBytes("ISO-8859-1");
-                if (!new String(recordData, "ISO-8859-1").equals(element.value)) {
-                    throw new ImageWriteException(
-                            "Invalid record value, not ISO-8859-1");
+                /**
+                 * favor raw bytes over value. This allows callers to use their
+                 * own encoding of fields.
+                 */
+                final byte[] recordData;
+                if( element.getRawBytes() != null && element.getRawBytes().length > 0 ) {
```
kinow: Ah, I had a better look using the [git history](https://github.com/apache/commons-imaging/commit/df41d46a01618a04c8f411c775f7df7f17e2c5e5#diff-ac232ea300229a1690ebdfdfec87bc99), and `getRawBytes` returned the bytes returned from a `getBytes("ISO-8859-1")` call.

image

So the only way that `getRawBytes` would return `null`, would be if there was an error encoding the String into ISO-8859-1. Which would also happen in the other branch of the `if` statement here.

So I believe the change wouldn't help you, as we would be still using ISO-8859-1 before you got the bytes.

garydgregory:
This needs a test.

PR #5, opened 2015-02-21
Proposal for a patch for Imaging-159 using a POJO

mgmechanics:
Proposal for a patch for Imaging-159 using a POJO.

I already applied the POJO Parameter Object classes so that they are replace the Map<String, Object> in all classes including test code.


src/main/java/org/apache/commons/imaging/ImageParser.java
```diff
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
```
britter: This has to be removed from all files. Please file an [ICLA](http://www.apache.org/licenses/).

src/main/java/org/apache/commons/imaging/ImageParser.java
```diff
@@ -954,13 +938,12 @@ protected final boolean canAcceptExtension(final String filename) {
...
-    protected BufferedImageFactory getBufferedImageFactory(final Map<String, Object> params) {
-        if (params == null) {
+    protected BufferedImageFactory getBufferedImageFactory(final ImagingParameters params) {
+        if (params == null || params.isBufferedImageFactoryPresent() == false) {
             return new SimpleBufferedImageFactory();
         }
 
-        final BufferedImageFactory result = (BufferedImageFactory) params
-                .get(ImagingConstants.BUFFERED_IMAGE_FACTORY);
+        final BufferedImageFactory result = params.getBufferedImageFactory();
```
britter: As noted in JIRA, to me it feels strange to pass a BufferedImageFactory via a parameters object. But since this has been the case before this PR, I think we should first concentrate on getting the ImagingParameters in place and refactor the BufferedImageFactory later (if that is possible)

src/main/java/org/apache/commons/imaging/Imaging.java
```diff
@@ -197,6 +193,8 @@ public static boolean hasImageFileExtension(String filename) {
      * @return An ImageFormat, such as ImageFormat.IMAGE_FORMAT_JPEG. Returns
      *         ImageFormat.IMAGE_FORMAT_UNKNOWN if the image type cannot be
      *         determined.
+     * @throws org.apache.commons.imaging.ImageReadException
```
britter: Looks like an unrelated change to me. Please keep change sets as minimal as possible. It makes reviewing easier.

src/main/java/org/apache/commons/imaging/Imaging.java
```diff
@@ -411,12 +417,14 @@ public static ICC_Profile getICCProfile(final InputStream is, final String filen
      * @param filename
      *            Filename associated with image data (optional).
      * @param params
-     *            Map of optional parameters, defined in ImagingConstants.
+     *            Optional parameters, defined in ImagingParameters.
```
britter: This is something we should work on after we have added ImagingParameters. IMHO if the parameter is optional, there should be a method overload that doesn't take the parameter.

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+ * Created 2015 by Michael Gross, mgmechanics@mgmechanics.de
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
+ */
+
+package org.apache.commons.imaging;
+
+import org.apache.commons.imaging.common.BufferedImageFactory;
+
+/**
+ * This class is a POJO holding data for parameters as requested in IMAGING-159.
```
britter: Strange comment :-) "as requested in IMAGING-159" can be dropped

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+ * Other parameters have a default value. This is told in the javadoc for their
+ * getX() method. They don't have a isXPresent() method. You may access them any time.
+ * <p>
+ * All boolean parameters have default parameters. Their setter and getter have a different
+ * naming scheme: enableX() causes isXEnabled() to return {@code true} while 
+ * disableX() causes isXEnabled() to return {@code false}.
+ */
+public class ImagingParameters {
+    
+    private boolean verbose; //PARAM_KEY_VERBOSE
```
britter: No need for this comments

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * This class is a POJO holding data for parameters as requested in IMAGING-159.
+ * It holds data needed for all image formats. For data needed only by one
+ * particular format there are inherited classes.
+ * <p>
+ * There are two kinds of parameters: with and without default value. For
+ * parameters without default value it implements lazy initialization:
+ * <br>
+ * After getting an instance there is no value present. In this state accessing
+ * the value with getX() in this state will cause a RuntimeException.
```
britter: Typo: one "in this state" can be dropped

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+ * There is always a isXPresent() method for this parameter which returns
+ * {@code false} in this state.
+ * <br>
+ * Once you provided a value using setX() you may access this value with getX().
+ * The isXPresent() method will return {@code true} now.
+ * <p>
+ * Other parameters have a default value. This is told in the javadoc for their
+ * getX() method. They don't have a isXPresent() method. You may access them any time.
+ * <p>
+ * All boolean parameters have default parameters. Their setter and getter have a different
```
britter: "... have default values"

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+     */
+    public boolean isVerboseEnabled() {
+        return this.verbose;
+    }
+    
+    /**
+     * Turns the verbose mode on.
+     * Parameter applies to read and write operations.
+     */
+    public void enableVerbose() {
```
britter: Would be better to add a plain old setter method so that this class can be used as a JavaBean.

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+    
+    //****** check methods ******
+    
+    /**
+     * Throws a RuntimeException if a given value is {code null}.
+     * @param value 
+     */
+    final void checkIfValueIsNull(final Object value) {
+        if (value == null) {
+            throw new IllegalArgumentException("The value for any parameter must not null.");
```
britter: Would be helpful for users to know which parameter caused the exception.

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+    
+    /**
+     * Throws a RuntimeException if the value for this parameter isn't set yet.
+     * @param value 
+     */
+    final void checkIfParameterIsPresent(final Object value) {
+        if (value == null) {
+            throw new IllegalStateException("You tried to get a value which is not present.");
```
britter: See above

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+}
```
britter: Please add a test case new code you introduce.

src/main/java/org/apache/commons/imaging/ImagingParameters.java
mgmechanics: On 23.02.2015 19:59, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/ImagingParameters.java:
      > +    /**
> +     * Tells if the verbose mode is turned on/off.
> +     * <p>
> +     * Parameter applies to read and write operations.
> +     * @return Valid values: {@code true} means {@literal "verbose mode enabled"}
> +     * and {@code false} means {@literal "verbose mode disabled"} (default value).
> +     */
> +    public boolean isVerboseEnabled() {
> +        return this.verbose;
> +    }
> +    
> +    /**
> +     * Turns the verbose mode on.
> +     * Parameter applies to read and write operations.
> +     */
> +    public void enableVerbose() {

      Would be better to add a plain old setter method so that this
        class can be used as a JavaBean.
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    > so that this class can be used as a JavaBean.
    
    You already can! All boolean parameters have a default value. So
    they are initialized when the class is instanziated.
    
    I added enable/disable methods for boolean values with this scenario
    in my mind: After releasing 1.0 we need to add a nuance of strict:
    Be a little bit strict.
    
    enum Strict {
        FULL,
        MEDIUM,
        LIBERAL
    }
    
    enableStrict() => Strict.FULL
    disableStrict() => Strict.LIBERAL
    enableStrictMedium() => Strict.MEDIUM
    
    There is no need to break the API.
    -- 
Michael Groß

src/main/java/org/apache/commons/imaging/ImagingParameters.java
mgmechanics: On 23.02.2015 20:00, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/ImagingParameters.java:
      > +     * @see org.apache.commons.imaging.PixelDensity
> +     */
> +    public void setPixelDensity(final PixelDensity value) {
> +        checkIfValueIsNull(value);
> +        this.pixelDensity = value;
> +    }
> +    
> +    //****** check methods ******
> +    
> +    /**
> +     * Throws a RuntimeException if a given value is {code null}.
> +     * @param value 
> +     */
> +    final void checkIfValueIsNull(final Object value) {
> +        if (value == null) {
> +            throw new IllegalArgumentException("The value for any parameter must not null.");

      Would be helpful for users to know which parameter caused the
        exception.
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    Added it :-)
    
    -- 
Michael Groß

src/main/java/org/apache/commons/imaging/ImagingParameters.java
mgmechanics: On 23.02.2015 20:02, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/ImagingParameters.java:
      > +    final void checkIfValueIsNull(final Object value) {
> +        if (value == null) {
> +            throw new IllegalArgumentException("The value for any parameter must not null.");
> +        }
> +    }
> +    
> +    /**
> +     * Throws a RuntimeException if the value for this parameter isn't set yet.
> +     * @param value 
> +     */
> +    final void checkIfParameterIsPresent(final Object value) {
> +        if (value == null) {
> +            throw new IllegalStateException("You tried to get a value which is not present.");
> +        }
> +    }
> +}

      Please add a test case new code you introduce.
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    
    Added tests.
    -- 
Michael Groß

src/main/java/org/apache/commons/imaging/ImagingParameters.java
mgmechanics: On 23.02.2015 19:59, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/ImagingParameters.java:
      > +    /**
> +     * Tells if the verbose mode is turned on/off.
> +     * <p>
> +     * Parameter applies to read and write operations.
> +     * @return Valid values: {@code true} means {@literal "verbose mode enabled"}
> +     * and {@code false} means {@literal "verbose mode disabled"} (default value).
> +     */
> +    public boolean isVerboseEnabled() {
> +        return this.verbose;
> +    }
> +    
> +    /**
> +     * Turns the verbose mode on.
> +     * Parameter applies to read and write operations.
> +     */
> +    public void enableVerbose() {

      Would be better to add a plain old setter method so that this
        class can be used as a JavaBean.
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    > so that this class can be used as a JavaBean
    
    Yesterday I wrote "You already can! All boolean parameters have a
    default value. So they are initialized when the class is
    instanziated.". I was a little bit fast.  Of course, if we need a
    full JavaBean we need to have plain old setter methods even for
    boolean values (for values of all other types there is one). So I
    just implemented plain setters.
    
    What is the intended use case for a JavaBean here? 
    -- 
Michael Groß

kinow:
Looks like the code from the pull request and the code in the master branch have moved into too different directions, and it can't be updated any longer.

@mgmechanics let me know if you are able to update your pull request to the latest code, please.

Thanks
Bruno

src/main/java/org/apache/commons/imaging/formats/pnm/PnmImageParser.java
```diff
@@ -334,6 +325,12 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
                     writer = new PamWriter();
                 }
             }
+            
+            // read parameters specific for the PNM format
+            if (params instanceof ImagingParametersPnm) {
```
kinow: Would be nice if we could use generics, instead of checking instance types. The parameters of `PnmImageParser` must definitely be an `ImagingParametersPnm` I believe.

src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
```diff
@@ -513,25 +497,36 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
...
-                } else if (text instanceof PngText.Itxt) {
-                    writeChunkiTXt(os, (PngText.Itxt) text);
-                } else {
-                    throw new ImageWriteException(
-                            "Unknown text to embed in PNG: " + text);
+        
+        // text chunks are parameters specific for PNG images
+        // so we need to ask first if parameters specific for PNG images are provided
+        // than ask for text chunks
+        if (parameters instanceof ImagingParametersPng) {
```
kinow: Would be good if generics could be used here to avoid the `instanceof`, or another way to model objects.

src/main/java/org/apache/commons/imaging/ImagingParametersPcx.java
```diff
@@ -0,0 +1,104 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.imaging;
+
+/**
+ * This class is a POJO holding data for parameters as requested in IMAGING-159.
+ * It holds additional data needed for the PCX format only.
```
kinow: Could be rewritten I think. Not necessary to say that it was requested and include what issue was.

src/main/java/org/apache/commons/imaging/ImagingParameters.java
```diff
@@ -0,0 +1,310 @@
...
+        String value = this.fileNameHint;
+        checkIfParameterIsPresent(value);
+        return value;
+    }
+    
+    /**
+     * Parameter used to hint the filename when reading from a byte array
+     * or InputStream. The filename hint can help disambiguate what file the
+     * image format.
+     * <p>
```
garydgregory, line 149: Close HTML tags.

garydgregory:
Needs a rebase and pre-review to make sure this is needed: Git master has an `ImagingParameters` today.

PR #8, opened 2015-02-21
Imaging-164

mgmechanics:
Patch for Imaging-164.


src/main/java/org/apache/commons/imaging/formats/ico/IcoImageParser.java
```diff
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
```
britter: Can not be applied with this change.

kinow:
Conflicts fixed manually, commits stashed and changes.xml entry added. Thanks again @mgmechanics !

PR #9, opened 2015-02-21
Imaging-163

mgmechanics:
Patch for Imaging-163. I replaced the abstract method by a concrete one establish returnin g null as a default behaviour if not implemented otherwise. Returning null was already found as default behaviour.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/ImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/ImageParser.java
@@ -13,8 +13,6 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
- * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
  */
 package org.apache.commons.imaging;
 
--- a/src/main/java/org/apache/commons/imaging/ImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/ImageParser.java
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
  */
 package org.apache.commons.imaging;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/apache/commons/imaging/ImageParser.java
```diff
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
```
britter: Can not be applied with this change.

src/main/java/org/apache/commons/imaging/ImageParser.java
```diff
@@ -682,8 +688,10 @@ public abstract Dimension getImageSize(ByteSource byteSource, Map<String, Object
      *                            parser implementation.
      * @throws IOException        In the event of unsuccessful read or access operation.
      */
-    public abstract String getXmpXml(ByteSource byteSource, Map<String, Object> params)
-            throws ImageReadException, IOException;
+    public String getXmpXml(final ByteSource byteSource, final Map<String, Object> params)
```
britter: I'm not sure whether it is a good idea to have this method here if it doesn't make sense for all formats (as stated in the JavaDoc)

mgmechanics: On 23.02.2015 20:15, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/ImageParser.java:
      > @@ -682,8 +688,10 @@ public abstract Dimension getImageSize(ByteSource byteSource, Map<String, Object
>       *                            parser implementation.
>       * @throws IOException        In the event of unsuccessful read or access operation.
>       */
> -    public abstract String getXmpXml(ByteSource byteSource, Map<String, Object> params)
> -            throws ImageReadException, IOException;
> +    public String getXmpXml(final ByteSource byteSource, final Map<String, Object> params)

      I'm not sure whether it is a good idea to have this method here
        if it doesn't make sense for all formats (as stated in the
        JavaDoc)
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    I found an abstract method and two ways to implement it:
    a. return some String != null
    b. return null
    
    Of course I did not touch all implementations of type (a). It seems
    that no all formats supply xmp xml strings. I made this patch to
    remove the need for implementations of type (b). Now everyone knows:
    If my implementation can not return any xmp xml it should return
    null. This was my intention. If we need another default return value
    than null - no problem.
    -- 
Michael Groß

britter: My feeling is, that we have a general design flaw here. If we have a a super class A and to inheriting class B and C, it makes no sense to define methods in A which are only supported by either B oder C but not both. If a method is added to A that doesn't make sense for either B or C then A doesn't seem to be the generalization of B and C.

kinow: Reading the issue, and this last comment, I agree about the issue being actually a bad design. IMO, we should move that method to an interface, perhaps ``. Then implementations can choose to implement it when appropriate.

So for `PcxImageParser` it would simply not implement it.

kinow:
Closed in #54

PR #11, opened 2015-02-21
Imaging-162

mgmechanics:
Proposed patch for Imaging-162.


src/main/java/org/apache/commons/imaging/formats/bmp/BmpHeaderInfo.java
```diff
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
```
britter: Can not be applied with this change.

src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
```diff
@@ -68,6 +69,26 @@
...
+     * With this constructor you can define the direction.
+     * @param readBottomUp {@code true}: the image data are stored bottom-up,
+     * {@code false} the image data are stored top-down
+     * @param width the width of the image to be built
+     * @param height the height of the image to be built
+     * @param hasAlpha indicates whether the image has an alpha channel
+     * (the selection of alpha channel does not change the memory
+     * requirements for the ImageBuilder or resulting BufferedImage.
+     */
+    public ImageBuilder(final boolean readBottomUp, final int width, final int height, final boolean hasAlpha) {
```
britter: Is there a way for us to detect whether the image is bottom up?

mgmechanics: On 23.02.2015 20:43, Benedikt Ritter
      wrote:
    
    
      In src/main/java/org/apache/commons/imaging/common/ImageBuilder.java:
      > +     * Construct an ImageBuilder instance.
> +     * <p>
> +     * Usually image data is stored bottom-up, that means that the last row in the
> +     * image data is the first row of pixels in the image as displayed.
> +     * Sometimes image data are stored top-down: The first row in the image data
> +     * is also the first row of pixels in the image as displayed.
> +     * With this constructor you can define the direction.
> +     * @param readBottomUp {@code true}: the image data are stored bottom-up,
> +     * {@code false} the image data are stored top-down
> +     * @param width the width of the image to be built
> +     * @param height the height of the image to be built
> +     * @param hasAlpha indicates whether the image has an alpha channel
> +     * (the selection of alpha channel does not change the memory
> +     * requirements for the ImageBuilder or resulting BufferedImage.
> +     */
> +    public ImageBuilder(final boolean readBottomUp, final int width, final int height, final boolean hasAlpha) {

      Is there a way for us to detect whether the image is bottom up?
      —
          it on GitHub.
      
        
          
          
        
        
      
    
    > Is there a way for us to detect whether the image is bottom up?
    
    I found a much simpler way without modifying ImageBuilder. Now it is
    only a matter inside org.apache.commons.imaging.formats.bmp package.
    For bmp image files: If the header has a positive height, it is
    bottom-up. You may read BitmapHeaderInfo.isBottomUpBitmap: If it is
    true the the height was positive and the bitmap is to read
    bottom-up, if false the height was negative and it is to read
    top-down.
    
    Before the patch the BitmapHeaderInfo.height may be negative, now it
    is always positive.
    
    I read the specs and docs again: Miroslav Gloub is right: If the
    image data are compressed, the height must be positive.
    -- 
Michael Groß

PR #12, opened 2015-02-22
IMAGING-165

mgmechanics:
I did something useful with unused field: added it to TiffContents


src/main/java/org/apache/commons/imaging/formats/tiff/TiffContents.java
```diff
@@ -13,6 +13,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Changed 2015 by Michael Gross, mgmechanics@mgmechanics.de
```
britter: Can not be applied with this change.

kinow:
Fixed conflicts manually, added changes.xml. Merged, thanks @mgmechanics !

PR #13, opened 2015-02-24
Imaging-134

mgmechanics:
At time I can only throw a proper exception. RST markers are valid markers in entropy data so the  exception message observed in IMAGING-97 and IMAGING-134 was misleading. I can not add the missing feature yet. But I added DRI and RST markers to the JpegConstants and JpegUtils class.


kinow:
Closed by #47 

Thanks!

PR #19, opened 2015-11-27
IMAGING-176

GabiAxel:
When "Compression" field is missing in TIFF - default to no compression to avoid exception.


britter:
Is the compression field optional in TIFF files or does the absence of the compression field indicate a broken file?

GabiAxel:
I'm not an expert on this, but according to https://en.wikipedia.org/wiki/Tagged_Image_File_Format#TIFF_Compression_Tag "The default value is 1 = no compression", which seems to me that the field is optional.

I got several such files from a user of my app, and they can be opened with different programs just fine, but cause an exception when attempting to get their info with Commons Imaging.

britter:
@GabiAxel okay! Do you see a way to provide a unit test for the new behavior?

GabiAxel:
I'll try to work on this in the next few days.

britter:
Awesome! Thank you.

GabiAxel:
@britter I pushed another change and a test TIFF file which caused failure in TiffReadTest and the change makes the test pass.

britter:
@GabiAxel I'll have a look tomorrow!

GabiAxel:
@britter `TiffReadTest` does that - it loads all TIFF images, and by adding the new image `src/test/data/images/tiff/8/no-compression-tag.tiff` this test runs `Imaging.getMetadata()` for it too. I didn't add new behaviour, just fixed how `Imaging.getMetadata()` deals with a case of an image, so I think it's enough.

britter:
Fixed in SVN in revision 1749362. Should sync to GitHub shortly. Thank you!

PR #27, opened 2017-07-13
Increase code coverage one

TheRealHaui:
I have created Unit Tests to increase code coverage which I want to contribute.


garydgregory:
Where some of these generated?

garydgregory:
@TheRealHaui:

Thank you for your PR. 

This is a lot of code, I think it would be appropriate for you to file a ICLA with us. Please see  https://www.apache.org/licenses/#clas and https://www.apache.org/licenses/icla.pdf

Thank you,
Gary

TheRealHaui:
@garydgregory 
You're welcome.
Where to send the pdf to?

garydgregory:
Please see https://www.apache.org/licenses/#submitting

TheRealHaui:
@garydgregory 
Thanks for your response.
Looked over the secretary@ section ...

Referring to your question if some of the tests were generated:
I use a lot of tools on top of my IDE that advise and suggest me where and how to improve code.
Some - like the Sonar plugin give advice concerning code improvements.
Some like code coverage tools advice me which lines/branches of code are not covered through automated tests yet.
And to those I sticked in the end here.

TheRealHaui:
Just submitted my ICLA to secretary@apache.org.

garydgregory:
Under what name? Feel free to email me at ggregory@apache.org if you do not want to state your name here.

TheRealHaui:
Just sent it to you via email.

garydgregory:
Did you try a build like "mvn clean verify"?

This PR format seems bogus as it contains folders like "org.apache.commons.imaging.common.itu_t4" instead "org/apache/commons/imaging/common/itu_t4"

TheRealHaui:
Well, now I did a "mvn clean verify".
It didn't complain about anything than for too less code coverage.

However, I've checked what you claimed referring the "directory structure".
You're right.
That's quite or better very odd.
I don't know how this happened right now exactly.
But - of course - I'll fix it.

TheRealHaui:
@garydgregory 
Here we go, directory naming problem solved.

TheRealHaui:
The CI fails due to the GrayscaleRountripTest which wasn't changed by me.
As it runs locally on my side I consider the timeout problem issued by the Travis server to be a local problem of the Travis installation in conjunction with the \@Theory test that gets executed.

src/test/java/org/apache/commons/imaging/formats/tiff/fieldtypes/FieldTypeByteTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+      FieldTypeByte fieldTypeByte = FieldType.UNDEFINED;
+      ByteOrder byteOrder = ByteOrder.BIG_ENDIAN;
+
+      try { 
+        fieldTypeByte.writeData( null, byteOrder);
+        fail("Expecting exception: Exception");
+      } catch(Exception e) {
+         assertEquals("Invalid data: null (null)",e.getMessage());
+         assertEquals(FieldTypeByte.class.getName(), e.getStackTrace()[0].getClassName());
```
onealj: It may be worth adding a helper method for this..

assertExceptionClassName(FieldTypeByte.class, e);

Why are you needing to look at the stack trace? Can't you look at the exception class itself?

src/test/java/org/apache/commons/imaging/formats/bmp/BmpWriterRgbTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.awt.image.BufferedImage;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for class {@link BmpWriterRgb}.
+ *
+ * @date 13.07.2017
```
TheRealHaui: Good idea.
Done.

TheRealHaui:
As the pull request is already closed shall I reopen another with the same name?

garydgregory:
Yes, you want a fresh PR.

Gary

src/test/java/org/apache/commons/imaging/formats/bmp/BmpWriterRgbTest.java
```diff
@@ -0,0 +1,35 @@
...
+    public void testGetImageData() {
+
+        BmpWriterRgb bmpWriterRgb = new BmpWriterRgb();
+        BufferedImage bufferedImage = new BufferedImage(2, 2, 5);
+        byte[] byteArray = bmpWriterRgb.getImageData(bufferedImage);
+
+        assertEquals( 24, bmpWriterRgb.getBitsPerPixel() );
+        assertEquals( 0, bmpWriterRgb.getPaletteSize() );
+        assertEquals( 16, byteArray.length );
+        assertEquals( "[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]", Arrays.toString(byteArray) );
```
TheRealHaui: Done.

src/test/java/org/apache/commons/imaging/common/bytesource/ByteSourceTest.java
```diff
@@ -66,6 +64,22 @@ protected File createTempFile(final byte src[]) throws IOException {
         }
         final byte longArray[] = (baos.toByteArray());
 
-        return new byte[][] { emptyArray, single, simple, zeroes, longArray, };
+        return new byte[][]{emptyArray, single, simple, zeroes, longArray,};
+    }
+
+    @Test
```
TheRealHaui: Yes, I know.
However, for what reason ever I experienced a lot of people preferring the try/catch notation over the annotation approach.
That's why I write/wrote it this way.
In the hope to have to rewrite as less code in the end as possible.

src/test/java/org/apache/commons/imaging/formats/jpeg/segments/JfifSegmentTest.java
```diff
@@ -0,0 +1,35 @@
...
+  @Test
+  public void testCreatesJfifSegment() {
+
+      byte[] byteArray = new byte[25];
+      JfifSegment jfifSegment = null;
+
+      try {
+        jfifSegment = new JfifSegment((-2275), byteArray);
+        fail("Expecting exception: Exception");
+      } catch(Throwable e) {
```
TheRealHaui: Had to do with the fact that at the beginning the test class resided in a different hierarchical folder structure than the original class ...
Corrected.

src/test/java/org/apache/commons/imaging/formats/pnm/PbmFileInfoTest.java
```diff
@@ -0,0 +1,41 @@
...
+ **/
+public class PbmFileInfoTest{
+
+
+  @Test
+  public void testGetRGBThrowsIOException() throws IOException {
+
+      PbmFileInfo pbmFileInfo = new PbmFileInfo(2764, 354, true);
+      byte[] byteArray = new byte[2];
+      ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArray);
```
TheRealHaui: Changed to interface.

src/test/java/org/apache/commons/imaging/formats/pnm/PbmFileInfoTest.java
```diff
@@ -0,0 +1,41 @@
...
+      byteArrayInputStream.read(byteArray);
+
+      try { 
+        pbmFileInfo.getRGB((InputStream) byteArrayInputStream);
```
TheRealHaui: Changed.

src/test/java/org/apache/commons/imaging/formats/tiff/fieldtypes/FieldTypeByteTest.java
TheRealHaui: You're right.
Omitting the stack trace is definitely shorter and more intuitive.

TheRealHaui:
Improvements made in #28.

src/test/java/org/apache/commons/imaging/common/bytesource/ByteSourceTest.java
jbduncan: @onealj @TheRealHaui If I were maintaining this code, I'd be happy with either pattern.

`@Test(expected=NullPointerException.class)` would be a little shorter and equivalent to the `try..catch` construct used here in this case.

However, the reason people tend to prefer `try..catch` is because in a lot of cases it is not granular enough. Take this (admittedly contrived) example:

```java
@Test(expected=NullPointerException.class)
public void doSomething() {
  String s = null;
  assertTrue(s.isEmpty());
  try {
    somethingWhichThrowsNullPointerException():
  } catch (NullPointerException e) {
    assertNull(s);
  }
}
```

In this case, rather than `somethingWhichThrowsNullPointerException(s1)` throwing `NullPointerException` (which we expect), it will be `s.isEmpty()` that throws it instead, which isn't what we intended, since it means the test will always "pass".

src/test/java/org/apache/commons/imaging/formats/tiff/fieldtypes/FieldTypeByteTest.java
TheRealHaui: No, you're not right!
Obviously, became already dizzy ...

The difference is that the approach with the stack trace - of course - verifies which class the exception originated from.
The verification which exception was thrown and before more over if an exception was thrown gets already verified in the catch clause.

PR #28, opened 2017-07-17
Increase code coverage one

TheRealHaui:
Added various improvements to selected Unit Tests.

PR #29, opened 2017-07-19
Improve various unit tests style

TheRealHaui:
Code formatting changed as desired in #27 and #28.


jbduncan:
Looks much better, thanks @TheRealHaui!

As I discussed in #28, I'm sure there's area for improvement on the whitespace/readability front, but that would require running `mvn checkstyle:check` and fixing the warnings it reports in a separate PR.

On a related note, Travis seems to complaining that GreyScaleRountripTest (unrelated to this PR AFAICT) killed the JVM process during testing. Perhaps Travis just needs to be restarted and the test will run properly? But I don't immediately know how one would restart Travis...

TheRealHaui:
You're welcome.
However, have to get used to the importance of style topics yet ...

Regarding the GreyScaleRoundTripTest:
Yes.
However, I thing or better guess it has to do with the amount of memory the Travis instance got supplied with.
As its a free service I'd await it to have very less memory.
And the specific test needs on my local machine quit a lot of memory as there is a JUnit theory inside.
Therefore I guess a restart won't help.
However again, I can only guess and I'am sure only owners can play around with CI instance settings.

PR #30, opened 2017-07-31
add-tests

TheRealHaui:
Added new Unit Tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,6 +209,12 @@
       <version>2.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>nl.jqno.equalsverifier</groupId>
+      <artifactId>equalsverifier</artifactId>
+      <version>2.3.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <reporting>
--- a/src/test/java/org/apache/commons/imaging/formats/jpeg/xmp/JpegXmpParserTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/jpeg/xmp/JpegXmpParserTest.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.apache.commons.imaging.formats.jpeg.xmp;
 
 import org.junit.Test;
--- a/src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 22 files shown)


TheRealHaui:
Checks failed due to Travis JDK crashing not due to problems in code.

pom.xml
```diff
@@ -209,6 +209,12 @@
       <version>2.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
kinow: -0 Looks like it reduces the code that needs to be written a little bit. But I'd still be keen to keep the code lengthier, but using just vanilla JUnit.

src/test/java/org/apache/commons/imaging/formats/jpeg/xmp/JpegXmpParserTest.java
```diff
@@ -1,3 +1,19 @@
+/*
```
kinow: Great catch on the missing licence headers! As I'm adding comments to this PR, and I'm just passing by the component to see if it supports something that I need for a project - and am not really familiar with the code base. It may take a bit longer to merge this PR. But if you submit another one just for the headers, we can merge it pretty much immediately ;-)

src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.commons.imaging.formats.pnm;
+
+import org.apache.commons.imaging.ImageReadException;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for class {@link PgmFileInfo}.
+ *
+ * @date 2017-08-01
```
kinow: Not necessary I think...

src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
```diff
@@ -0,0 +1,49 @@
...
+ * @see PgmFileInfo
+ *
+ **/
+public class PgmFileInfoTest{
+
+  @Test(expected = ImageReadException.class)
```
kinow: I think the IDE got confused here. Better to keep consistent with number of spaces.

src/test/java/org/apache/commons/imaging/formats/tiff/taginfos/TagInfoByteOrShortTest.java
```diff
@@ -0,0 +1,55 @@
...
+      TagInfoByteOrShort tagInfoByteOrShort = new TagInfoByteOrShort("r", 500, 500, tiffDirectoryType);
+      ByteOrder byteOrder = ByteOrder.BIG_ENDIAN;
+      short[] shortArray = new short[2];
+      byte[] byteArray = tagInfoByteOrShort.encodeValue(byteOrder, shortArray);
+
+      assertArrayEquals(new byte[] {(byte)0, (byte)0, (byte)0, (byte)0}, byteArray);
+  }
+
+  @Test
+  public void testEncodeValueTaking1And1AndEncodeValueTaking1And1AndEncodeValueTaking1And1ReturningNonEmptyArrayTwo() {
```
kinow: Never had problems with really long test method names, but **if** it's possible to simplify the method name, I'd be +1 for that. Not thoroughly reading the tests yet.

src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
TheRealHaui: Removed the class comment.

src/test/java/org/apache/commons/imaging/formats/pnm/PgmFileInfoTest.java
TheRealHaui: Reformatted code.

src/test/java/org/apache/commons/imaging/formats/tiff/taginfos/TagInfoByteOrShortTest.java
TheRealHaui: Renamed test methods.

pom.xml
TheRealHaui: Well, Equalsverifier checks equals and hashCode methods soroly.
Therefore I considered to use it.

TheRealHaui:
Changes incorporated.
Build working.

pom.xml
kinow: Indeed, and I may even use it in some projects at work. However, other commons components use - as far as I can remember - JUnit with some local helper classes for tests. So my line of thinking, is that even though it looks useful, others would have to learn and remember to use it, or eventually we would end up with calls to Equalsverifier methods, and in some other places the old way. But again, -0 is not a blocker, if others prefer to have it, I'm fine with it :-)

TheRealHaui: Understand your point.

src/test/java/org/apache/commons/imaging/formats/jpeg/xmp/JpegXmpParserTest.java
TheRealHaui: Created #31 which contains only license header corrections.

kinow:
Looking good. I'm still just reluctant to add the new test dependency. So leaving it here for now until others have time to chime in.

I'm updating merge requests now, merging what can be merged, so that we can start modifying code for issues found when discussing a possible 1.x release.

PR #32, opened 2017-10-15
add test for DataReaderStripsTest::applyPredictor

testingsavvy:
This PR adds a new test for `DataReaderStripsTest::applyPredictor`, so that it can be tested in isolation.


kinow:
Added an extra commit to include the license header.

Thanks for your contribution @testingsavvy . Pull request merged.

PR #33, opened 2017-12-04
Imaging-209 PnmImageParser throws ImageWriteException when writing

aesy:
if PnmImageParser.PARAM_KEY_PNM_RAWBITS is provided.

I fixed the issue and added a test to cover it.

Issue can be found here: https://issues.apache.org/jira/browse/IMAGING-209


kinow:
Thank you **a lot** for including tests, and making the change so simple to be reviewed. No extra formatting, or any other change :-) 

Had a look at the code in Eclipse and also at the JIRA description, and I agree it makes sense that entry from the map. Merged!

PR #34, opened 2017-12-07
IMAGING-188

bad-member:
https://issues.apache.org/jira/browse/IMAGING-188

commits: 
- support for all icns types listed on wikipedia: https://en.wikipedia.org/wiki/Apple_Icon_Image_format
- support for JPEG 2000 via modified ImageIO Library, look at the new entry in the pom

- test for the enhanced IcnsDecoder, plus test files (extracts images of all icns-types from several icns files)


DasBueroAmDraht:
The tests run successfully if they are run separately, but not by Jenkins. Thats why it's failing. Please review.

pom.xml
```diff
@@ -209,6 +209,12 @@
       <version>2.6</version>
       <scope>test</scope>
     </dependency>
+    <!-- necessary modification of ImageIO for support of JPEG 2000 -->
+    <dependency>
+      <groupId>com.github.jai-imageio</groupId>
+      <artifactId>jai-imageio-jpeg2000</artifactId>
+      <version>1.3.0</version>
+    </dependency>
```
kinow: Adding a dependency to `jai-imageio-jpeg2000` brings `jai-imageio-core` too I believe. ImageIO and Commons Imaging share a lot of features. But, IMO, adding a dependency from one to another, would be like adding TestNG as a dependency in JUnit to get the TestNG groups feature.

A better approach, IMO, would be to build up the required code for JPEG2000 either within Commons Imaging or as a module. And then implement the rest of the code as you have it here.

garydgregory:
Closing: No reply.

PR #35, opened 2018-02-05
IMAGING-154 Removed Debug class

kinow:
Removed the `Debug` and `ImageDump` classes. Also removed methods with the `verbose` parameter, as well as any `System.out.println`'s. Tried to keep the change to a minimum, but given the extent of its use, I believe the pull request diff will be a bit long.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
@@ -19,7 +19,6 @@ package org.apache.commons.imaging.formats.bmp;
 import static org.apache.commons.imaging.ImagingConstants.BUFFERED_IMAGE_FACTORY;
 import static org.apache.commons.imaging.ImagingConstants.PARAM_KEY_FORMAT;
 import static org.apache.commons.imaging.ImagingConstants.PARAM_KEY_PIXEL_DENSITY;
-import static org.apache.commons.imaging.ImagingConstants.PARAM_KEY_VERBOSE;
 import static org.apache.commons.imaging.common.BinaryFunctions.read2Bytes;
 import static org.apache.commons.imaging.common.BinaryFunctions.read4Bytes;
 import static org.apache.commons.imaging.common.BinaryFunctions.readByte;
@@ -37,6 +36,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import org.apache.commons.imaging.FormatCompliance;
 import org.apache.commons.imaging.ImageFormat;
@@ -54,6 +55,9 @@ import org.apache.commons.imaging.palette.PaletteFactory;
 import org.apache.commons.imaging.palette.SimplePalette;
 
 public class BmpImageParser extends ImageParser {
+
+    private static final Logger LOGGER = Logger.getLogger(BmpImageParser.class.getName());
+
     private static final String DEFAULT_EXTENSION = ".bmp";
     private static final String[] ACCEPTED_EXTENSIONS = { DEFAULT_EXTENSION, };
     private static final byte[] BMP_HEADER_SIGNATURE = { 0x42, 0x4d, };
@@ -90,7 +94,7 @@ public class BmpImageParser extends ImageParser {
     }
 
     private BmpHeaderInfo readBmpHeaderInfo(final InputStream is,
-            final FormatCompliance formatCompliance, final boolean verbose)
+            final FormatCompliance formatCompliance)
             throws ImageReadException, IOException {
         final byte identifier1 = readByte("Identifier1", is, "Not a Valid BMP File");
         final byte identifier2 = readByte("Identifier2", is, "Not a Valid BMP File");
@@ -184,7 +188,7 @@ public class BmpImageParser extends ImageParser {
             throw new ImageReadException("Invalid/unsupported BMP file");
         }
 
-        if (verbose) {
+        if (LOGGER.isLoggable(Level.FINE)) {
             debugNumber("identifier1", identifier1, 1);
             debugNumber("identifier2", identifier2, 1);
             debugNumber("fileSize", fileSize, 4);
@@ -300,16 +304,16 @@ public class BmpImageParser extends ImageParser {
     }
 
     private BmpImageContents readImageContents(final InputStream is,
-            final FormatCompliance formatCompliance, final boolean verbose)
+            final FormatCompliance formatCompliance)
             throws ImageReadException, IOException {
-        final BmpHeaderInfo bhi = readBmpHeaderInfo(is, formatCompliance, verbose);
+        final BmpHeaderInfo bhi = readBmpHeaderInfo(is, formatCompliance);
 
         int colorTableSize = bhi.colorsUsed;
         if (colorTableSize == 0) {
             colorTableSize = (1 << bhi.bitsPerPixel);
         }
 
-        if (verbose) {
+        if (LOGGER.isLoggable(Level.FINE)) {
             debugNumber("ColorsUsed", bhi.colorsUsed, 4);
             debugNumber("BitsPerPixel", bhi.bitsPerPixel, 4);
             debugNumber("ColorTableSize", colorTableSize, 4);
@@ -326,8 +330,8 @@ public class BmpImageParser extends ImageParser {
 
         switch (bhi.compression) {
         case BI_RGB:
-            if (verbose) {
-                System.out.println("Compression: BI_RGB");
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine("Compression: BI_RGB");
             }
             if (bhi.bitsPerPixel <= 8) {
                 paletteLength = 4 * colorTableSize;
@@ -341,8 +345,8 @@ public class BmpImageParser extends ImageParser {
             break;
 
         case BI_RLE4:
-            if (verbose) {
-                System.out.println("Compression: BI_RLE4");
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine("Compression: BI_RLE4");
             }
             paletteLength = 4 * colorTableSize;
             rleSamplesPerByte = 2;
@@ -353,8 +357,8 @@ public class BmpImageParser extends ImageParser {
             break;
         //
         case BI_RLE8:
-            if (verbose) {
-                System.out.println("Compression: BI_RLE8");
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine("Compression: BI_RLE8");
             }
             paletteLength = 4 * colorTableSize;
             rleSamplesPerByte = 1;
@@ -365,8 +369,8 @@ public class BmpImageParser extends ImageParser {
             break;
         //
         case BI_BITFIELDS:
-            if (verbose) {
-                System.out.println("Compression: BI_BITFIELDS");
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine("Compression: BI_BITFIELDS");
             }
             if (bhi.bitsPerPixel <= 8) {
                 paletteLength = 4 * colorTableSize;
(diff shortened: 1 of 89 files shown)


kinow:
`changes.xml` updated. Last commit contains the word-trick to close the pull request. Once merged, everything should be correct, just need to update JIRA.

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
```diff
@@ -327,7 +331,7 @@ private BmpImageContents readImageContents(final InputStream is,
         switch (bhi.compression) {
         case BI_RGB:
             if (verbose) {
```
garydgregory: Seeing "if verbose" and then "log at the info level" is slightly confusing here. If we want to keep the 'verbose' setting then maybe the code should rename this to 'logInfo', or, should we provide a setting that contains the log level to use when 'verbose' is enabled?

src/main/java/org/apache/commons/imaging/formats/png/chunks/PngChunkText.java
```diff
@@ -42,9 +47,9 @@ public PngChunkText(final int length, final int chunkType, final int crc, final
         final int textLength = bytes.length - (index + 1);
         text = new String(bytes, index + 1, textLength, StandardCharsets.ISO_8859_1);
 
-        if (getDebug()) {
-            System.out.println("Keyword: " + keyword);
-            System.out.println("Text: " + text);
+        if (LOGGER.isLoggable(Level.FINE)) {
```
garydgregory: Uhg, the JUL level names are AWFUL :=(

src/main/java/org/apache/commons/imaging/formats/psd/PsdHeaderInfo.java
```diff
@@ -45,9 +50,14 @@ public PsdHeaderInfo(final int version, final byte[] reserved, final int channel
     }
 
     public void dump() {
-        final PrintWriter pw = new PrintWriter(new OutputStreamWriter(System.out, Charset.defaultCharset()));
-        dump(pw);
-        pw.flush();
+        try (StringWriter sw = new StringWriter(); PrintWriter pw = new PrintWriter(sw)) {
```
garydgregory: This seems like duplicate code from a similar block I saw above. Should it be refactored?

src/main/java/org/apache/commons/imaging/formats/psd/PsdImageContents.java
```diff
@@ -40,9 +45,14 @@ public PsdImageContents(final PsdHeaderInfo header,
     }
 
     public void dump() {
-        final PrintWriter pw = new PrintWriter(new OutputStreamWriter(System.out, Charset.defaultCharset()));
-        dump(pw);
-        pw.flush();
+        try (StringWriter sw = new StringWriter(); PrintWriter pw = new PrintWriter(sw)) {
```
garydgregory: Same comment as above.

src/main/java/org/apache/commons/imaging/formats/tiff/TiffField.java
```diff
@@ -355,9 +360,14 @@ private String getValueDescription(final Object o) {
     }
 
     public void dump() {
-        final PrintWriter pw = new PrintWriter(new OutputStreamWriter(System.out, Charset.defaultCharset()));
-        dump(pw);
-        pw.flush();
+        try (StringWriter sw = new StringWriter(); PrintWriter pw = new PrintWriter(sw)) {
```
garydgregory: Same comment as above.

src/main/java/org/apache/commons/imaging/icc/IccTag.java
```diff
@@ -73,11 +76,12 @@ private IccTagDataType getIccTagDataType(final int quad) {
     }
 
     public void dump(final String prefix) throws ImageReadException, IOException {
-        final PrintWriter pw = new PrintWriter(new OutputStreamWriter(System.out, Charset.defaultCharset()));
-
-        dump(pw, prefix);
-
-        pw.flush();
+        try (StringWriter sw = new StringWriter(); PrintWriter pw = new PrintWriter(sw)) {
```
garydgregory: Same comment as above even though this code is slightly different - no catch clause.

src/test/java/org/apache/commons/imaging/formats/jpeg/exif/WriteExifMetadataExampleTest.java
```diff
@@ -27,7 +27,7 @@
 import org.apache.commons.imaging.formats.jpeg.JpegImageParser;
 import org.apache.commons.imaging.formats.tiff.TiffField;
 import org.apache.commons.imaging.formats.tiff.TiffImageMetadata;
-import org.apache.commons.imaging.util.Debug;
+import org.apache.commons.imaging.internal.Debug;
```
garydgregory: I like the package name :-)

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
sebbASF: Seems to me it would be better to use the log-level alone to control what is being logged.

So the verbose flag would become a log-level instead.

src/main/java/org/apache/commons/imaging/formats/png/chunks/PngChunkText.java
kinow: I went straight for `LOGGER.isDebugEnabled()` and the IDE remembered me what it actually looked like :=/

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
kinow: You are both correct @garydgregory , @sebbASF .

I tried to remove the flags `DEBUG` or `isDebug()` or `getDebug()` by `LOGGER.isLoggable(Level.FINE)`, but forgot to do the same for `verbose` using the `INFO` level (wonder if we should use FINEST for debug, and FINE for the verbose messages...)
Will just log to info. I believe we can assume that

src/main/java/org/apache/commons/imaging/icc/IccTag.java
kinow: This was intentional.

The previous code contains the `throws IOException` (while the other methods don't) -> https://github.com/apache/commons-imaging/blob/d2ec76bd10f30c39ae5180ede1254908e76045f0/src/main/java/org/apache/commons/imaging/icc/IccTag.java#L75

Should we make them all the same, catching and logging, instead of throwing the IOException? As this is a `dump` method, not used by another method such as `toString()`, I think the impact on end users should be minimal.

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
kinow: Done in https://github.com/apache/commons-imaging/pull/35/commits/96b0a18e3c0ac435da85a3cf361975e2e77aa7f0

Also removed the `PARAM_KEY_VERBOSE` constant, used to include a param in a map to enable verbose. I initially put a `@deprecated` annotation, then realized this will be 1.0 of imaging, not sanselan, so removing it should be OK.

kinow: And on

>(wonder if we should use FINEST for debug, and FINE for the verbose messages...)

If you have the chance to run the tests, you will notice the huge amount of verbose messages being printed now. I believe some users could be inundated with log lines (as the INFO is enabled by default? and verbose was disabled by default)

garydgregory:
In general I do not think a library should log at INFO, only DEBUG or TRACE.

sebbASF:
Agreed

kinow:
Agreed on not using INFO level @sebbASF & @garydgregory .

Moved what was in FINE to FINEST level. And moved what was in INFO to FINE.

My plan for the web site information is to have the following information, just with more complete information:

>we are using jul
>
>exceptions will be logged to SEVERE
>a more verbose output is available at FINE
>nformation for debug and troubleshooting is available at FINEST

@garydgregory addressed your comments about code duplication with comments above.

If there are no other objections here or in the mailing list, I am planning to merge this work next weekend (that should give plenty of time for others to review, and time for me to try to work on the **last** blocking issue for 1.0: documentation! 

Thanks!

andreasrosdal:
I hope we can get a 1.0 release of Commons Imaging very soon. Thanks for working towards this!

kinow:
Thanks @andreasrosdal ! I still have some more development bandwidth this month, so hopefully it will be merged this weekend, and I will find a way to squeeze in a few more hours to finish the website updates. Then we will be ready to call a 1.0 vote again.

@garydgregory ping. I left a few replies to your comments. I think that'd be OK to merge as is, unless you prefer to address the exception handling & code duplication now.

garydgregory:
+1 to merge. I'd prefer to see a 1.0-alpha1 or 1.0-beta1 released next instead of 1.0 but I'm not RM'ing this one... ;-)

kinow:
Thanks @garydgregory ! No objection for a 1.0-alpha1, then wait a couple of months or a bit more perhaps for user feedback. I assume this 1.0-alpha1 would make all the way to maven central?

andreasrosdal:
I hope that you will publish the release to the maven central repos, since I hope to use it as a dependency.

kinow:
>So, what is the code migration path for ImagingConstants.PARAM_KEY_VERBOSE? In other words, how should we turn on some sort of file parser debug logging?

We are using `java.util.logging`, JUL. You can tweak the logging level to see the messages that were previously written to the stdout when the `PARAM_KEY_VERBOSE` was enabled.

https://commons.apache.org/proper/commons-imaging/gettingstarted.html (see Logging)

>I am asking because every now and then we have to deal with non‑compliant JPEG files and we would like to get some clue or provide some info to the user that a file is broken.

Could you try the JUL and let us know if it contains enough info to tell users what is wrong when reading or parsing the JPEG files, please? If there's anything lacking, feel free to open an issue and maybe point at what could be improved (e.g. number of segments, stage of parsing where it failed parsing, etc.)

kinow:
>So, for our purposes, should we set the log Level to Level.FINE on the loggers named "org.apache.commons.imaging", "org.apache.commons.imaging.common", and "org.apache.commons.imaging.formats.jpeg"? Specifying exact class names would be way too granular (and generate too much source code) for us.

JUL loggers are [hierarchical](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html),

>A Logger object is used to log messages for a specific system or application component. Loggers are normally named, using a hierarchical dot-separated namespace. Logger names can be arbitrary strings, but they should normally be based on the package name or class name of the logged component, such as java.net or javax.swing. (...)

If you enable for `org.apache.commons.imaging` then you will enable all the loggers. If you would like to limit FINE log output to a package like jpeg, you can do that too like your example for `formats.jpeg`.

>We are basically interested in getting a hint at where the parser encounters an error in the binary file data stream. In other words, where the file starts deviating from the specification, even when the error is recoverable. We would like to get a specification compliance hint. We do not want or need strict specification compliance enforcement.

:+1: 

>I am sorry for not being able to describe more precisely what we would like to see from Commons Imaging. Anyway, thank you for you support. +1

Not a problem at all. I think this is what most users of the library want. In case it helps, what I normally do is try to parse the images with Apache Commons Imaging, then if I get an error, before even trying to tweak the log levels or debugging, I test it with `exiftool` and maybe GIMP. IIRC the command I normally use it `exiftool dump --html $img > output.html && firefox output.html` ? Or something similar that gives me the complete image data separated by segments, metadata, byte information, etc.

exiftool is quite good at pointing whether the image is corrupt or if there's some structural errors.

If exiftool looks good, but Imaging is not able to parse it, then it could be because we are being too strict regarding the standard (exiftool is a bit more lenient), or we are missing to handle something specified in an Adobe spec extension, or some color space that's not supported/handled, etc.

In case you are able to identify that we are missing some logging you can always create a JIRA issue :slightly_smiling_face: just try to attach the test image, code snippet, and as much info as possible.

Cheers

PR #36, opened 2018-03-28
Avoid jpeg parse fail due to jfxx and exif

hyunuck:
- add jfxx signature when parsing jpeg
- Avoid jpeg parse fail due to exif

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
@@ -312,7 +312,12 @@ public class JpegImageParser extends ImageParser {
     @Override
     public ImageMetadata getMetadata(final ByteSource byteSource, final Map<String, Object> params)
             throws ImageReadException, IOException {
-        final TiffImageMetadata exif = getExifMetadata(byteSource, params);
+        TiffImageMetadata exif = null;
+        try {
+            exif = getExifMetadata(byteSource, params);
+        } catch (Exception e) {
+            // exif parse failed.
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
```diff
@@ -312,7 +312,12 @@ private void dumpSegments(final List<? extends Segment> v) {
     @Override
     public ImageMetadata getMetadata(final ByteSource byteSource, final Map<String, Object> params)
             throws ImageReadException, IOException {
-        final TiffImageMetadata exif = getExifMetadata(byteSource, params);
+        TiffImageMetadata exif = null;
+        try {
+            exif = getExifMetadata(byteSource, params);
+        } catch (Exception e) {
+            // exif parse failed.
+        }
```
kinow: I think we should still fail here. Other parsers such as the `TiffImageParser` fail, so `JpegImageParser` not throwing an exception would be wrong.

Also, not even logging the exception can cause problems to users of the API that would have to debug the code to understand why they were not seeing the EXIF metadata, just to find an exception that was not propagated.

The rest of the code looks good. Would you be able to provide at least one unit tests for your code @hyunuck, please?

Thanks
Bruno

kinow: Ping

garydgregory: Ping AND there are no tests.

PR #37, opened 2018-04-20
[IMAGING-248]: ICNS: missing element types; some safety checks

gregsh:
The problematic  incs file: groovy-1.7.5/bin/groovy.icns

From https://bintray.com/groovy/maven/groovy/1.7.5#files

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/icns/IcnsDecoder.java
+++ b/src/main/java/org/apache/commons/imaging/formats/icns/IcnsDecoder.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.commons.imaging.ImageReadException;
+import org.apache.commons.imaging.Imaging;
 import org.apache.commons.imaging.common.ImageBuilder;
 import org.apache.commons.imaging.formats.icns.IcnsImageParser.IcnsElement;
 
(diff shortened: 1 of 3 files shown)


gregsh:
We've started using commons-imaging for indexing image meta info in IntelliJ IDEA and related IDEs since 2018.1 and come across several previously unsupported cases in ICNS reader.

Xcode.icns (from IntelliJ codebase testData) with PNG and raw BMP images as PNG
groovy.icns (from groovy-1.7.5.zip) which is completely broken

As much as I love tests I'm not sure how to proceed. So what do you think?

kinow:
>As much as I love tests I'm not sure how to proceed. So what do you think?

Maybe it we could get at least one or two tests, maybe using that `groovy-1.7.5/bin/groovy.icns` file? That'd be enough to include the new code IMO. But without any tests, either myself or another developer would have to write tests for it later, which would be much harder.

Right now my main focus is getting everything sorted for the 1.0 release. If we can't get tests for this, I would still like to use this contribution, but would move it to 1.1 or later, until I have time to write the unit tests :+1:

src/main/java/org/apache/commons/imaging/formats/icns/IcnsDecoder.java
```diff
@@ -21,6 +21,7 @@
 import java.util.List;
 
 import org.apache.commons.imaging.ImageReadException;
+import org.apache.commons.imaging.Imaging;
```
kinow: This could lead to a circular dependency between the package `oac.imaging`, and the type `Imaging`, with this file. So I will try to use the parser directly instead 👍

src/main/java/org/apache/commons/imaging/formats/icns/IcnsImageParser.java
```diff
@@ -195,19 +195,14 @@ public void dump(final PrintWriter pw) {
...
-                                                                                  // bytes,
-                                                                                  // including
-                                                                                  // this
-                                                                                  // header
+    private IcnsElement readIcnsElement(final InputStream is, final int remainingSize) throws IOException {
+        // Icon type (4 bytes)
+        final int type = read4Bytes("Type", is, "Not a Valid ICNS File", getByteOrder());
+        // Length of data (4 bytes), in bytes, including this header
+        final int elementSize = read4Bytes("ElementSize", is, "Not a Valid ICNS File", getByteOrder());
+        if (elementSize > remainingSize) {
```
kinow: The `groovy.icns` file mentioned in the PR fails to parse here 🤔 resulting in an exception

kinow:
Tried converting it to PNG with `icns2png`, but it also failed:

```
$ icns2png -x groovy.icns 
----------------------------------------------------
Reading icns family from groovy.icns...
libicns: icns_read_family_from_file: Error reading icns file - all parsing methods failed!
Unable to read icns family from file groovy.icns!
Errors while extracting icns data from groovy.icns!
```

kinow:
There was a commit 9 months ago to fix the ICNS ico. Tested this one from `master` and it was successfully parsed: https://github.com/apache/groovy/blob/cbb338854ee1e39a819575ad8ed0f87925a17d73/src/bin/groovy.icns

kinow:
There was a commit 9 months ago to fix the ICNS ico. Tested this one from `master` and it was successfully parsed: https://github.com/apache/groovy/blob/cbb338854ee1e39a819575ad8ed0f87925a17d73/src/bin/groovy.icns

kinow:
Added a unit test to parse the `groovy.icns` file from the project `master` branch, and also to parse the `python.icns` (it contains an ic09 image). Ran the test against `master`, and it failed as the number of images did not match.

To get the right number of images, I compared the output of this branch with the output of `icns2png` (both matched! :tada:).

kinow:
Merged after adding the test and fixing a few checkstyle issues. Thanks for the PR @gregsh !

PR #40, opened 2019-01-02
IMAGING-220: replace while by if in the JPEG decoder extend method

kinow:
Looking at the specification and other implementations (see JIRA ticket for links), it appears to me that the `while` in the `extend` method is actually supposed to be an `if`. 

All tests passing after this change, also added an extra test to validate an existing bug found by Guido Vranken, where the decoder would take longer than 10 seconds to process the file. This issue does not apply to Sanselan, only to Commons Imaging.

Will merge in the next days if there are no objections.

Cheers
Bruno


kinow:
Merged in master

PR #43, opened 2019-05-29
[IMAGING-239] Add inflate (deflate algorithm) to TIFF files

pauldaustin:
- Adds inflate (deflate algorithm) to TIFF files.
- Also allows passing params to reading directories, so that image data loading can be turned off
- Make getTiffRawImageData public so that the image data can be loaded after the directories.
 
I was going to do 2 commits but then it got committed by accident.


kinow:
From an initial look the code looks good. @pauldaustin could you create a JIRA issue for it, and also include some basic unit tests, please? There's some info on contributing [here](https://github.com/apache/commons-imaging/blob/master/CONTRIBUTING.md). Thanks!

src/main/java/org/apache/commons/imaging/common/Deflate.java
```diff
@@ -0,0 +1,40 @@
...
+    throws ImageReadException {
+    try {
+      final Inflater inflater = new Inflater();
+      inflater.setInput(bytes);
+      final byte[] result = new byte[expected];
+      inflater.inflate(result);
+      return result;
+    } catch (final DataFormatException e) {
+      throw new ImageReadException("Unable to decompress image", e);
+    }
```
kinow: Formatting needs to be changed to 4 spaces to match other classes.

src/main/java/org/apache/commons/imaging/common/Deflate.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.imaging.common;
+
+import java.util.zip.DataFormatException;
+import java.util.zip.Inflater;
+
+import org.apache.commons.imaging.ImageReadException;
+
+public class Deflate {
```
kinow: Probably some basic javadocs could help other devs/users :+1:

src/main/java/org/apache/commons/imaging/formats/tiff/TiffReader.java
```diff
@@ -412,7 +423,7 @@ public void read(final ByteSource byteSource, final Map<String, Object> params,
         readDirectories(byteSource, formatCompliance, listener);
     }
 
-    private TiffImageData getTiffRawImageData(final ByteSource byteSource,
+    public TiffImageData getTiffRawImageData(final ByteSource byteSource,
```
kinow: Making methods that were private now public always need a short explanation, as these methods will be now part of the maintained API accessible by users, and we will have to have more care to change it.

Could you explain why these two methods in TiffReader are public and not private, please?

pauldaustin: I reverted these changes

kinow:
Thanks for working on the PR comments so quickly. Could you take a look at the Travis CI broken build? It appears to be an issue with the code formatting.

You can look at the .travis.yml file, and run the same maven targets in your own computer to reproduce it.

pauldaustin:
Do you know if there are an exported configurations for Eclipse code style, formatting and import ordering. I use that on my own projects so it automatically gets formatted properly when I save.

I avoided using the eclipse auto formatter which would resolve these issues as it re-formatted a bunch of the other code. My code style also sorts methods into a consistent order which if run regularly on a project makes merging changes between branches easier.

pauldaustin:
Looks like the formatter wouldn't have helped as the spaces were at the end of lines in the javadoc comment and it didn't strip them when I tried the formatter.

kinow:
>Looks like the formatter wouldn't have helped as the spaces were at the end of lines in the javadoc comment and it didn't strip them when I tried the formatter.

:+1: I kind of got used to the style used in most Commons projects, so I don't have an imported style/formatter settings for Eclipse (I also use it for Java). Some projects have mixed styles too, mixing old and new, with different rules. Easier to try to copy the existing style and run Maven targets before committing I guess.

New error:

```
[ERROR] 7002: org.apache.commons.imaging.common.bytesource.ByteSource: Method 'public java.lang.String getFilename()' has been removed
```

I think an existing method that was public has now been removed? That is not backward compatible - i.e. existing users would have issues unless we released a new major version (next release is still maintenance/alpha).

garydgregory:
I renamed getFilename() (obvious typo) to getFileName(). Since we do not have a 1.0 release yet, breaking BC is OK. We should get the public names right before 1.0.

kinow:
>I renamed getFilename() (obvious typo) to getFileName(). Since we do not have a 1.0 release yet, breaking BC is OK. We should get the public names right before 1.0.

Is there a way to tell cliir to ignore these errors for now then? So that the PR's checks can all pass

pauldaustin:
I merged master into my branch to get the other changes.

Maybe this will fix the PR check issues.

kinow:
@pauldaustin no that won't fix the build. I've created a PR that disables Clirr check for backward compatibility - https://github.com/apache/commons-imaging/pull/45

If/once that gets merged, if you rebase your changes again then the build should pass this time.

Thanks
Bruno

kinow:
Done @pauldaustin . If you rebase this PR onto master, the tests should finally pass. Sorry for the delay!

pauldaustin:
I'm giving up, seems to be a bunch of errors I'm not sure how to fix, or are in parts of the code I haven't changed.

kinow:
@pauldaustin just had a look, and the PR is passing in some JDK's, and the failures appear to be related to Travis-CI. Nothing wrong with your code as far as I can tell, so I believe it's ready to be reviewed and hopefully merged :) 👍 thanks for rebasing it

pauldaustin:
@kinow That's great news thanks.

kinow:
Merged, thanks!

PR #44, opened 2019-06-03
[IMAGING-228] Drop cube method and add unit tests for PhotometricInterpretationLogLuv

kinow:
Noticed that the `PhotometricInterpretaterLogLuv` had no coverage, so decided to write some simple unit tests.

Then found a `cube` method that can be replaced by `Math.pow(N, 3)`, and also that there is no javadocs about what is XYZ, why LAB, what is CIE, or even what is Photometric Interpretation.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -45,6 +45,9 @@ The <action> type attribute can be add,update,fix,remove.
   </properties>
   <body>
     <release version="1.0-alpha2" date="2019-??-??" description="Second 1.0 alpha release">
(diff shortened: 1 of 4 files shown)


kinow:
Just need to finish the unit tests to promote to PR. Draft for now.

kinow:
`oraclejdk8` is failing, but looks to be related to https://travis-ci.community/t/the-travis-ci-build-could-not-be-completed-due-to-an-error/1345

kinow:
Addressed feedback on using double instead of integer on `Math.pow`.

PR #45, opened 2019-06-04
Disable clirr report for now (alpha releases)

kinow:
There are pull requests which Travis jobs are failing due to clirr check goal failures.

We have renamed methods `getFilename` to `getFileName`, which breaks backward compatibility. That's, however, acceptable given we are working on an 1.0-alpha release series.

So I suggest we disable in Travis-CI Clirr checks for now, by simply removing `clirr:check` from the Maven default goal in the project `pom.xml`.

Cheers
Bruno

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,7 @@
   <inceptionYear>2007</inceptionYear>
 
   <build>
-    <defaultGoal>clean verify apache-rat:check clirr:check checkstyle:check findbugs:check javadoc:javadoc</defaultGoal>
+    <defaultGoal>clean verify apache-rat:check checkstyle:check findbugs:check javadoc:javadoc</defaultGoal>
     <plugins>
       <plugin>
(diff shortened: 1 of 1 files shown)


kinow:
If merged, I will create a ticket in JIRA, set to 1.0 release of imaging, that says to re-enable the clirr reports once the full release is out :+1:

kinow:
A new PR by another contributor got marked as failed in Travis. Merging this and asking him to rebase. There's some ongoing discussion in the mailing list about alpha/beta releases, that involves this component. But in case we decided for keeping BC, then we would just have to re-enable clirr, and undo the commit that broke BC (though I believe we will keep using alpha here as right now).

kinow:
Issue created to re-enable it later: https://issues.apache.org/jira/browse/IMAGING-229

With pointers to the two e-mail threads related to this issue.

PR #51, opened 2019-08-17
IMAGING-210: prevent NegativeArraySizeException in chunks in PngImageParser

kinow:
Fix similar to IMAGING-211. The unit test is actually in the same file.

Simply checks immediately the length of a chunk. If the parser finds a chuck with a negative chunk, it exits immediately with an `ImageReadException`, instead of failing with other exceptions.


kinow:
The only failure in Travis was due a Javadoc issue with Java 13 (see #52). Will merge this one and continue working on a fix for javadoc in the separate issue.

PR #53, opened 2019-09-03
Close OutputStream before toByteArray() on underlying ByteArrayOutputStream

emopers:
When an `OutputStream` instance wraps an underlying `ByteArrayOutputStream` instance,
it is [recommended](https://stackoverflow.com/questions/2984538/how-to-use-bytearrayoutputstream-and-dataoutputstream-simultaneously-java) to flush or close the `OutputStream` before invoking the underlying instances' `toByteArray()`.

This pull request adds a call to `close` at the end of the `write` method of `TiffImageWriterLossy`.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterLossy.java
+++ b/src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterLossy.java
@@ -50,6 +50,8 @@ public class TiffImageWriterLossy extends TiffImageWriterBase {
         final BinaryOutputStream bos = new BinaryOutputStream(os, byteOrder);
 
         writeStep(bos, outputItems);
+
+        bos.close();
     }
 
     private void updateOffsetsStep(final List<TiffOutputItem> outputItems) {


src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterLossy.java
```diff
@@ -50,6 +50,8 @@ public void write(final OutputStream os, final TiffOutputSet outputSet)
         final BinaryOutputStream bos = new BinaryOutputStream(os, byteOrder);
 
         writeStep(bos, outputItems);
+
+        bos.close();
```
kinow: Hi @emopers 

You are correct on that. The issue on this method (and I think there might be other places in Imaging with the same issue) is that this was by design.

The `BinaryOutputStream` created here wraps the parameter `os`, an `OutputStream` passed by the user. We shouldn't close the stream that the user passed, but calling `BinaryOutputStream#close` will close the underlying stream as well - https://github.com/apache/commons-imaging/blob/8ee267db7a05d26f66e5f0b0ec7af6b813f99345/src/main/java/org/apache/commons/imaging/common/BinaryOutputStream.java#L71

That's why some time ago, when we fixed several cases of resources left open by using the `try-with-resources` trick, we left this one as-is.

kinow:
Closing as per comment above. I've added a comment (b684eea67f98010b67db9587dbfb0b1a2d02bb1a) to let other developers know that the resource is intentionally left open. Thanks!

PR #54, opened 2019-10-28
[IMAGING-163] Add XmpEmbedabble interface to parsers that support it

kinow:
An alternative propose to IMAGING-163. See #9 for another approach.

In this PR, a new interface `XmpEmbeddable` is added. The `getXmpXml` method is removed from the `ImageParser`, and from any parser that was always returning `null`.

Now, instead, only the parsers that currently support XMP metadata tags implement that method. Not all parsers will implement this interface, as not all image format specifications support XMP.


kinow:
Won't really merge it myself without another reviewer, first so that someone else helps deciding on which approach, and also because I am not really good with names.

PR #55, opened 2019-10-30
[IMAGING-236] Add support to read multiple images from GIF

christoffer-rydberg:
This PR adds support to read more than one image from a GIF file. It also allows reading of metadata needed for visualization (such as the offset for each image). I did not yet create a jira account but will do so for any future pull requests. I quickly dug into code as I got frustrated with an old [bug](https://stackoverflow.com/questions/22259714/arrayindexoutofboundsexception-4096-while-reading-gif-file) in ImageIO. I hope you'll accept my PR or let me know what to change to meet the code standard.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadata.java
+++ b/src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadata.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.imaging.formats.gif;
+
+import org.apache.commons.imaging.common.ImageMetadata;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+public class GifImageMetadata implements ImageMetadata {
+    private static final String NEWLINE = System.getProperty("line.separator");
+    private final int width;
+    private final int height;
+    private final List<GifImageMetadataItem> items;
+
+    GifImageMetadata(int width, int height, List<GifImageMetadataItem> items) {
+        this.width = width;
+        this.height = height;
+        this.items = Collections.unmodifiableList(new ArrayList<>(items));
+    }
+
+    @Override
+    public String toString(String prefix) {
+        prefix = prefix == null ? "" : prefix;
+        final StringBuilder result = new StringBuilder();
+        result.append(String.format("%sGIF metadata:", prefix));
+        result.append(String.format("%sWidth: %d%s", prefix, width, NEWLINE));
+        result.append(String.format("%sHeight: %d%s", prefix, height, NEWLINE));
+        result.append(NEWLINE);
+        result.append(String.format("%sImages:", prefix));
+        for (GifImageMetadataItem item : items) {
+            result.append(NEWLINE);
+            result.append(item.toString(prefix));
+        }
+        return result.toString();
+    }
+
+    public int getWidth() {
+        return width;
+    }
+
+    public int getHeight() {
+        return height;
+    }
+
+    @Override
+    public List<GifImageMetadataItem> getItems() {
+        return items;
+    }
+}
--- a/src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadata.java
+++ b/src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadata.java
@@ -22,22 +22,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-/*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 public class GifImageMetadata implements ImageMetadata {
     private static final String NEWLINE = System.getProperty("line.separator");
     private final int width;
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadata.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package org.apache.commons.imaging.formats.gif;
+
+import org.apache.commons.imaging.common.ImageMetadata;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/*
```
kinow: License duplicated?

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -25,7 +25,7 @@
 import static org.apache.commons.imaging.common.BinaryFunctions.readByte;
 import static org.apache.commons.imaging.common.BinaryFunctions.readBytes;
 
-import java.awt.Dimension;
+import java.awt.*;
```
kinow: I think this is one of the issues reported by Travis.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -492,17 +502,29 @@ public Dimension getImageSize(final ByteSource byteSource, final Map<String, Obj
             throw new ImageReadException("GIF: Couldn't read ImageDescriptor");
         }
 
-        // Prefer the size information in the ImageDescriptor; it is more
-        // reliable
-        // than the size information in the header.
```
kinow: We will need to understand why the old code was using the other width/height too.

kinow:
>I did not yet create a jira account but will do so for any future pull requests

We can create it later too, so no problem.

>I quickly dug into code as I got frustrated with an old bug in ImageIO. I hope you'll accept my PR or let me know what to change to meet the code standard.

My only concern is if this is not too specific to the GIF format. But will need to read the proposed change in the code later with more calm.

The code looks good, with comments and some tests. That's helpful for reviewers, thanks!!! Feel free to ping me if I take too long to look into this again 👍

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
kinow: Thanks for the description and link! Will read it when reviewing it again :tada:

christoffer-rydberg:
Thanks, I'll push an update to fix the mentioned issues.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
christoffer-rydberg: Added a comment in code and a unit test

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -485,24 +495,35 @@ public Dimension getImageSize(final ByteSource byteSource, final Map<String, Obj
...
-                IMAGE_SEPARATOR);
-        if (id == null) {
-            throw new ImageReadException("GIF: Couldn't read ImageDescriptor");
-        }
-
-        // Prefer the size information in the ImageDescriptor; it is more
-        // reliable
-        // than the size information in the header.
-        return new Dimension(id.imageWidth, id.imageHeight);
+
```
kinow: This looks fine for me, but Travis appears to be complaining about this line? :confused: 

```
[ERROR] src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java:[498] (regexp) RegexpSingleline: Line has trailing spaces.
```

christoffer-rydberg: Thanks, I saw. Had some trailing spaces. Fixed.

kinow:
Thanks @christoffer-rydberg, Travis build is happy. Just need to find spare time + :coffee: to have a play, but everything looks excellent so far :+1:

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.apache.commons.imaging.formats.gif;
+
+import org.apache.commons.imaging.common.ImageMetadata;
+
+public class GifImageMetadataItem implements ImageMetadata.ImageMetadataItem {
+    private static final String NEWLINE = System.getProperty("line.separator");
+    private final int delay;
+    private final int leftPosition;
+    private final int topPosition;
```
kinow: Maybe we should include the `width` and `height` of each gif image?

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -439,6 +439,16 @@ private GifBlock findBlock(final List<GifBlock> blocks, final int code) {
         return null;
     }
 
+    private <T extends GifBlock> List<T> findAllBlocks(final List<GifBlock> blocks, final int code) {
+        final List<T> filteredBlocks = new ArrayList<>();
+        for (final GifBlock gifBlock : blocks) {
+            if (gifBlock.blockCode == code) {
+                filteredBlocks.add((T) gifBlock);
```
kinow: Hmmm, this is tricky. We could have a malformed GIF that has a block with the wrong code, resulting in a runtime exception here I think? Or if I request something like `List<ImageDescriptor> descriptors = findAllBlocks(imageContents.blocks, GRAPHIC_CONTROL_EXTENSION);`. I wonder if we could come up with a different approach to avoid issues like this?

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
kinow: If possible. Just tried changing the code a bit, and I couldn't find the data for each image. Wondering if all we get is the left/top values instead?

kinow:
@christoffer-rydberg issue created https://issues.apache.org/jira/browse/IMAGING-236, will update the PR title

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
kinow: (it seems to be internal only, so maybe this should be fine for now)

kinow: Just re-read the `GifImageParser#readBlocks()`, and actually I believe the only way this could happen, is if the `switch/case` logic was wrong, and used the wrong class for a block code. So this should be fine, especially as this appears to be internal use only.

So maybe just add the suppress annotation and a note, e.g.

```
// See GifImageParser#parse for reference how the blocks are created. They should match
// the code we are giving here, returning the correct class type. Internal only.
@SuppressWarnings("unchecked")
private <T extends GifBlock> List<T> findAllBlocks(final List<GifBlock> blocks, final int code) {
```

?

kinow:
@christoffer-rydberg finished reviewing it. Had a quick look at the specification, then played with the code. Had added some comments yesterday on GitHub, then went through them this morning discarding most of them as I read the code :coffee: 

Left three, but I think only the one in the test file may need code change? Once you have had time to reply or resolve the comments/reviews, it can be merged and should be in 1.0-alpha2 :+1: thanks!!!

src/test/java/org/apache/commons/imaging/formats/gif/GifReadTest.java
```diff
@@ -57,10 +58,40 @@ public void testImageInfo() throws Exception {
...
+            final int xOffset = metadataItem.getLeftPosition();
+            final int yOffset = metadataItem.getTopPosition();
+            width = Math.max(width, image.getWidth() + xOffset);
+            height = Math.max(height, image.getHeight() + yOffset);
+        }
+
+        assertEquals(width, metadata.getWidth());
+        assertEquals(height, metadata.getHeight());
+        assertEquals(width, imageInfo.getWidth());
+        assertEquals(height, imageInfo.getHeight());
```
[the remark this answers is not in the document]
christoffer-rydberg: The first two asserts are for the metadata and the second two are for the image info. It's just to ensure the width/height is correctly set in both.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
christoffer-rydberg: I only use the metadata to decide the delay and placement for each image. For that I need the overall GIF width and height as well as the individual image offsets. As you say, you can easily get the width and height from each buffered image. Maybe it could be useful if you want to display such information before loading the actual images?

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
christoffer-rydberg: I think that sounds good. I'll add the comment and suppress the warning.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
christoffer-rydberg: Either we add it now just in case or we wait until we see there is an actual need for it? I'll let you decide! :) I can fix it if needed tomorrow.

src/test/java/org/apache/commons/imaging/formats/gif/GifReadTest.java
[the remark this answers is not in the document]
kinow: Aahh, just my eyes failing me. Sorry I thought it was the same object.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
kinow: Let's get this one merged in then, and add more later if necessary. Thanks!

PR #58, opened 2019-11-18
[IMAGING-243] PNG Writer Indexed Color with semi-transparent Pixels and Better Compression 

adorn:
I needed smaller PNG Files, so I updated the PNG Writer:

- Indexed Color with transparent Alpha / semi-transparent Pixels
- Better Compression

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
+++ b/src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
@@ -59,7 +59,7 @@ class PngWriter {
     }
 
     private void writeChunk(final OutputStream os, final ChunkType chunkType,
-                            final byte[] data) throws IOException {
+            final byte[] data) throws IOException {
         final int dataLength = data == null ? 0 : data.length;
         writeInt(os, dataLength);
         os.write(chunkType.array);
@@ -87,8 +87,8 @@ class PngWriter {
         public final InterlaceMethod interlaceMethod;
 
         ImageHeader(final int width, final int height, final byte bitDepth,
-                    final PngColorType pngColorType, final byte compressionMethod, final byte filterMethod,
-                    final InterlaceMethod interlaceMethod) {
+                final PngColorType pngColorType, final byte compressionMethod, final byte filterMethod,
+                final InterlaceMethod interlaceMethod) {
             this.width = width;
             this.height = height;
             this.bitDepth = bitDepth;
@@ -280,7 +280,7 @@ class PngWriter {
     }
 
     private void writeChunkSCAL(final OutputStream os, final double xUPP, final double yUPP, final byte units)
-            throws IOException {
+          throws IOException {
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
 
         // unit specifier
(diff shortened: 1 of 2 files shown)


adorn:
Thank you,  
JIRA issue https://issues.apache.org/jira/browse/IMAGING-243

src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
```diff
@@ -596,8 +588,10 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
             // Debug.debug("uncompressed", uncompressed.length);
 
             final ByteArrayOutputStream baos = new ByteArrayOutputStream();
-            final DeflaterOutputStream dos = new DeflaterOutputStream(baos);
             final int chunkSize = 256 * 1024;
+            Deflater deflater = new Deflater(Deflater.BEST_COMPRESSION);
```
kinow: Hi @adorn what if we keep the default behaviour (i.e. use default compression level) but allow users to change it?

See https://github.com/adorn/commons-imaging/pull/1

adorn: That's also OK.  Should I add some PngConstants?

PARAM_KEY_PNG_DEFLATE_COMPRESSION_LEVEL  ="PNG_DEFLATE_COMPRESSION_LEVEL";

    public static final int PNG_DEFLATE_COMPRESSION_LEVEL_DEFAULT= Deflater.DEFAULT_COMPRESSION;
    public static final int PNG_DEFLATE_COMPRESSION_LEVEL_BEST_COMPRESSION= Deflater.BEST_COMPRESSION;
    public static final int PNG_DEFLATE_COMPRESSION_LEVEL_BEST_SPEED= Deflater.BEST_SPEED;

kinow: I think it's not necessary. These values are already at the `Deflater` class, and may change. So we should be better allowing users to pass anything. There is also a ticket I think to allow for other compression algorithms (even though specification uses only Deflater, it says vendors can use other algorithms), in which case it could lead to having more values for each algorithm.

src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
```diff
@@ -396,6 +363,7 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
         }
         params.remove(ImagingConstants.PARAM_KEY_PIXEL_DENSITY);
         params.remove(PngConstants.PARAM_KEY_PHYSICAL_SCALE);
+        params.remove(PngConstants.PARAM_KEY_PNG_DEFLATE_COMPRESSION_LEVEL);
```
kinow: The current API around those parameters is a bit strange. It requires you to remove them all, in order to validate you have valid parameters. It definitely needs changing, but for now we will have to stick with the current API.

Looking at your change, you used `getOrDefault` instead of the suggested patch in that PR - which would work great. But you also had to call `.remove` here. So I think `compressionLevel` will be always-ever the default level?

adorn: Sorry didn't see your Pull Request

kinow:
Merged, thanks!

PR #59, opened 2019-11-19
[IMAGING-242] Upgrade to Junit5

kinow:
Used @mureinik 's excellent work on [this PR for Commons CSV](https://github.com/apache/commons-csv/pull/49) for the same end.

Only difference is that Commons Imaging uses Theories (`@Theory`, `@Datapoint`, etc). And this is [not available in JUnit 4](https://github.com/junit-team/junit5/pull/1422#issuecomment-389644868), so I used [JUnit Vintage](https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-vintage) as in [this post](https://www.igorski.co/java/junit/mixing-junit-4-and-junit-5-tests/).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,6 +225,7 @@
       <version>2.1</version>
       <scope>test</scope>
     </dependency>
+    <!-- the next three dependencies are for tests with theories/datapoints, see IMAGING-242 -->
     <dependency>
       <groupId>org.junit.platform</groupId>
       <artifactId>junit-platform-launcher</artifactId>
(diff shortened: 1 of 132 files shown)


mureinik:
Thanks for the shout-out @kinow !

I took a look and the repository after this PR was applied, and I think it's a pretty small delta to reach a Jupiter-only project, even if it does require implementing the "logic" of a Cartesian product of arguments - see my attempt at https://github.com/apache/commons-imaging/pull/61

The side advantage of this is to lock down the dependencies and not allow people to mix Jupiter and Vintage in the same class as was done in `TestImageReadException` and `TestImageWriteException` (presumably by mistake?).

kinow:
>The side advantage of this is to lock down the dependencies and not allow people to mix Jupiter and Vintage in the same class as was done in TestImageReadException and TestImageWriteException (presumably by mistake?).

Probably mea culpa. I did 90% with IDE search-replace following the docs & looking at your CSV PR. But for Theories, Datapoints and other things that weren't in CSV I struggled a bit. Thanks for tidying it up :+1:

PR #61, opened 2019-11-24
IMAGING-242 Complete the migration to JUnit Jupiter

mureinik:
This PR follows up https://github.com/apache/commons-imaging/pull/59 and completes the migration to JUnit Jupiter:

- Usages of `org.junit.Assert.assert[Not]Null` were replaced with `org.junit.jupiter.api.Assertions.asserrt[Not]Null`.
- Usages of `@Theories` and `@DataPoints` where replaces with JUnit Jupiter's `@ParameterizedTest` with `@MethodSources`. Unfortunately, Jupiter doesn't have a built-in capability to create a Cartesian product of datapoints like `@Theories` does, so this logic had to be explicitly implemented in `RountripBase`.
- With this cleanup, the additional Junit 5 Vintage dependencies could be removed.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/roundtrip/BitmapRoundtripTest.java
+++ b/src/test/java/org/apache/commons/imaging/roundtrip/BitmapRoundtripTest.java
@@ -18,16 +18,14 @@
 package org.apache.commons.imaging.roundtrip;
 
 import java.awt.image.BufferedImage;
+import java.util.stream.Stream;
 
-import org.junit.experimental.theories.DataPoints;
-import org.junit.experimental.theories.Theories;
-import org.junit.experimental.theories.Theory;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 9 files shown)


src/test/java/org/apache/commons/imaging/roundtrip/BitmapRoundtripTest.java
```diff
@@ -40,10 +38,12 @@
             TestImages.createBitmapBitmapImage(300, 300), // larger than 256
     };
 
-    @DataPoints
-    public static FormatInfo[] formatInfos = FormatInfo.READ_WRITE_FORMATS;
+    public static Stream<Arguments> testBitmapRoundtrip() {
```
kinow: Ah! `Arguments`? Hadn't seen this in their docs. Brilliant!

kinow:
Thanks a lot @mureinik ! I learned a bit more after this PR, and now Imaging is fully Junit 5 :confetti_ball:

PR #62, opened 2019-11-29
[IMAGING-245] Add disposal method to GIF metadata

christoffer-rydberg:
Add disposal method to GIF metadata in order to know how to render GIF frames with transparent pixels. See [here](http://webreference.com/content/studio/disposal.html) and [here](https://www.w3.org/Graphics/GIF/spec-gif89a.txt) for more info.


kinow:
Thanks for the PR. Travis seems to be failing, and we will need a JIRA for that too.

I am with limited internet until Sun/Mon, but can take a look at these issues later if you don't have time.

christoffer-rydberg:
Forgot to add license. Hope Travis will be happy now.

kinow:
Another checkstyle issue:

```
[ERROR] src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java:[47,47] (blocks) LeftCurly: '{' at column 47 should have line break after.
```

src/main/java/org/apache/commons/imaging/formats/gif/GifImageMetadataItem.java
```diff
@@ -42,13 +44,16 @@ public int getTopPosition() {
         return topPosition;
     }
 
+    public DisposalMethod getDisposalMethod() { return disposalMethod; }
```
kinow: Just format as other methods and checkstyle test should pass, probably Travis will give the green flag after this :+1:

christoffer-rydberg: Thanks. Fixed formatting.

christoffer-rydberg:
@kinow Any suggestions how to unit test this PR? Basically the only added logic is a private method to convert from internal gif int value to an enum. I could:
A) Make it public (bad since it is only needed by the parser).
B) Test it using reflection (not a fan)
C) Test it indirectly by adding more test gifs with different disposal methods (need to find such gifs).
D) Move the method to another class (e.g. MethodDisposalConverter) with a public testable method and inject it into the parser.

kinow:
Hi @christoffer-rydberg 

I still need to read the links and the code. But from your list of options, I'd be +1 for C), and +0 for D).

C would be my favourite as we would minimize changes to the API - unless there is some good reason anyway for this change, then we can discuss D too :+1:

christoffer-rydberg:
Thanks. A bit busy the next few days but I'll add a unit test when I get time over.

kinow:
Will try to review this one this week or next weekend. Thanks for updating it @christoffer-rydberg !

kinow:
Merged, thanks!

PR #64, opened 2019-12-14
IMAGING-151: use a copy of the list, make members package-protected/private, and re-use Comparator

kinow:
This PR creates a `Comparator` for `ColorCount`, replacing two comparators created in different parts of the code with the same logic.

The list that was sorted was a member of a class that could be used elsewhere. So instead of altering the values in that list, now other classes need to access that list through a method that returns a copy of the list.

Members in the package-protected class were made package-protected (from public) or private.

Bruno

### DIFF
--- a/src/main/java/org/apache/commons/imaging/palette/LongestAxisMedianCut.java
+++ b/src/main/java/org/apache/commons/imaging/palette/LongestAxisMedianCut.java
@@ -74,13 +74,14 @@ public class LongestAxisMedianCut implements MedianCut {
             }
         };
 
-        Collections.sort(colorGroup.colorCounts, comp);
+        final List<ColorCount> colorCounts = colorGroup.getColorCounts();
+        Collections.sort(colorCounts, comp);
         final int countHalf = (int) Math.round((double) colorGroup.totalPoints / 2);
         int oldCount = 0;
         int newCount = 0;
         int medianIndex;
-        for (medianIndex = 0; medianIndex < colorGroup.colorCounts.size(); medianIndex++) {
-            final ColorCount colorCount = colorGroup.colorCounts.get(medianIndex);
+        for (medianIndex = 0; medianIndex < colorCounts.size(); medianIndex++) {
+            final ColorCount colorCount = colorCounts.get(medianIndex);
 
             newCount += colorCount.count;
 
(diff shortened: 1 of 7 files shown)


kinow:
Added a few comments too, but planning to add more later when working on IMAGING-146.

src/main/java/org/apache/commons/imaging/palette/LongestAxisMedianCut.java
```diff
@@ -59,28 +59,14 @@ public boolean performNextMedianCut(final List<ColorGroup> colorGroups, final bo
...
-                case BLUE:
-                    return c1.blue - c2.blue;
-                default:
-                    return 0;
-            }
-        };
-
-        Collections.sort(colorGroup.colorCounts, comp);
+        final List<ColorCount> colorCounts = colorGroup.getColorCounts();
+        Collections.sort(colorCounts, new ColorCountComparator(mode));
```
kinow: Sorting the local variable now.

src/main/java/org/apache/commons/imaging/palette/MedianCutQuantizer.java
```diff
@@ -130,7 +130,7 @@ public Palette process(final BufferedImage image, final int maxColors,
 
             colorGroup.paletteIndex = i;
 
-            if (colorGroup.colorCounts.size() < 1) {
+            if (colorGroup.getColorCounts().isEmpty()) {
```
kinow: There are several other places with similar code smell, will raise a PR in the next hours to use `isEmpty()` in the rest of the code when possible.

src/main/java/org/apache/commons/imaging/palette/MostPopulatedBoxesMedianCut.java
```diff
@@ -52,13 +50,13 @@ public boolean performNextMedianCut(final List<ColorGroup> colorGroups,
             if (ignoreAlpha && colorComponent == ColorComponent.ALPHA) {
                 continue;
             }
-            Collections.sort(colorGroup.colorCounts, new ColorComparer(colorComponent));
+            Collections.sort(colorCounts, new ColorCountComparator(colorComponent));
```
kinow: This is sorted multiple times for each color component. Later a best color component will be chosen, and a final sort will take place.

PR #68, opened 2020-01-11
[IMAGING-246] Skip Photoshop IPTC blocks that are not recommended to be interpreted by parsers in the specification

kinow:
Specifically 1084, 1085, 1086, 1086 from https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/.

Unit test attached, using image provided in the JIRA issue.


src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
```diff
@@ -598,12 +598,12 @@ public JpegPhotoshopMetadata getPhotoshopMetadata(final ByteSource byteSource,
...
 
             final PhotoshopApp13Data data = segment.parsePhotoshopSegment(params);
-            if (data != null && photoshopApp13Data != null) {
-                throw new ImageReadException(
-                        "Jpeg contains more than one Photoshop App13 segment.");
+            if (data != null) {
+                if (photoshopApp13Data != null) {
+                    throw new ImageReadException("Jpeg contains more than one Photoshop App13 segment.");
+                }
+                photoshopApp13Data = data;
```
kinow: The logic removed had an issue, where if the first data was not null, then the second was null; the second value would be used. Resulting in missing photoshop APP13 data.

PR #73, opened 2020-04-09
IMAGING-253: ByteSourceInputStream has initialized its length when reading starts

b923:
The reason was that `streamLength` property was never initialized.

Also added test for verification


kinow:
Merged. Change looked good, included a test and no other failures. Formatted a bit the test file to include newline at the end of the file and use 4-spaces. Added a changes.xml and updated JIRA ticket too to include it in the next release. Thanks a lot @b923 !!!

b923:
Thank you @kinow

PR #75, opened 2020-05-17
Imaging 247

gwlucastrig:
This PR addresses JIRA issue 247.  Upon acceptance, I believe that 247 can be closed.


kinow:
Good explanation both here and on JIRA. Thanks again @gwlucastrig ! Change looks good to me!

src/main/java/org/apache/commons/imaging/formats/tiff/photometricinterpreters/PhotometricInterpreterPalette.java
```diff
@@ -45,11 +46,24 @@ public PhotometricInterpreterPalette(final int samplesPerPixel,
...
+        // This interpreter is used with TIFF_COMPRESSION_PACKBITS (32773).
+        // which unpacks to 8 bits per sample.  But if the bits-per-pixel
+        // is less than 8 bits, some authoring tools do not zero-out the
+        // unused bits.  This results in cases where the decoded by index
+        // exceeds the size of the palette.  So we set up a mask to protect
+        // the code from an array bounds exception.
+        int temp = 0;
+        for (int i = 0; i < bitsPerPixel; i++) {
+            temp = (temp << 1) | 1;
+        }
```
kinow, line 59: I liked the for-loop here! Much simpler than something like `int temp = (bitsPerPixel < 32 ? (1 << bitsPerPixel) -1 : -1)` which I **think** would give something similar, but much harder to maintain/read. :clap: we definitely need more clarity in our code, and your comment above is very helpful too!

PR #80, opened 2020-05-30
Fixed bug writing data to closed output

gwlucastrig:
A few years ago, somebody changed the ApacheImagingSpeedAndMemoryTest.java to close the output between test operations.  So when the application tried to report the results of the second timing test, the application would fail on an IOException "trying to write to closed output".

I wanted to fix the problem doing the minimum change I could.  So I modified the program to use a print stream rather than a formatter.  

Anyway, the changes should be minor and easily identified.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/examples/ApacheImagingSpeedAndMemoryTest.java
+++ b/src/test/java/org/apache/commons/imaging/examples/ApacheImagingSpeedAndMemoryTest.java
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 /****************************************************************
  * Explanation of Use and Rationale For Procedures
  *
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/imaging/examples/ApacheImagingSpeedAndMemoryTest.java
```diff
@@ -280,7 +261,6 @@ private void performTest(final String name) {
                 // this isn't fatal, but shouldn't happen
                 iex.printStackTrace();
             }
-
         }
     }
-}
+}
```
kinow, line 266: Missing git newline here I think?

kinow:
Looks good to me @gwlucastrig ! I'm on leave next week, and as it's winter over here, I might have time to prepare alpha2 release in the next days. Let me know if you have any other issues you'd like to have included in this release :+1: it will include your code for floating-points in tiff images.

gwlucastrig:
Thanks.

I am a quite busy with work right now and won't have time to work on Commons Imaging until late July.  I will eventually submit code changes for some of the JIRA items I submitted.  But I don't think you should delay alpha2 in order to wait for them.

PR #89, opened 2020-07-26
Replace FindBugs by SpotBugs, and suppress current issues

kinow:
The new dependabot issues are failing. The GitHub action is running SpotBugs, while Travis runs just `mvn`.

This PR updates the GitHub action to run `mvn`. Also replaces FindBugs by SpotBugs (using Lang as example), and suppresses the issues reported by SpotBugs so that build passes (there are issues in JIRA for that already).

`BinaryConstants` also implemented the `clone()` method without implementing `Cloneable`. This PR removes the `clone` method as that was not used.


kinow:
Interesting. When I use JDK 14 I get more errors, just like in Travis. Fixing those now.

kinow:
Getting some issues due to incompatibilities of SpotBugs with newer JVM versions, e.g. https://github.com/spotbugs/spotbugs/issues/756

Converted PR to draft. Let me see if I can fix the GitHub workflow. Otherwise will disable it and leave Travis only as I'm planning to cut a release soon.

kinow:
Java 16 error is due to FindBugs not supporting that JVM version.

PR #98, opened 2020-09-19
[IMAGING-265] ArrayIndexOutOfBoundsException on reading simple GeoTIFF

gwlucastrig:
This change addresses JIRA issue IMAGING-265.  It enhances the TIFF reader logic so that it can handle certain kinds of images that divide RGB values into three separate planes rather than using the more common interleaved format.

At this time, I have only implemented PlanarConfiguration=2 support for TIFF files organized into strips (rather than tiles) using the 24 bit RGB photometric interpreter.


gwlucastrig:
Pending review, I believe that this pull request addresses IMAGING-265 and that the issue can be closed.

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageData.java
```diff
@@ -140,8 +145,8 @@ public ImageDataReader getDataReader(final TiffDirectory directory,
     public abstract ImageDataReader getDataReader(TiffDirectory directory,
             PhotometricInterpreter photometricInterpreter, int bitsPerPixel,
             int[] bitsPerSample, int predictor, int samplesPerPixel, int width,
-            int height, int compression, ByteOrder byteOrder) throws IOException,
-            ImageReadException;
+      int height, int compression, int planarConfiguration,
+      ByteOrder byteOrder) throws IOException,            ImageReadException;
```
kinow, line 149: Not sure if this is GitHub's UI issue, but looks like there are several spaces between `IOException, ` and `ImageReadException`?

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
```diff
@@ -898,7 +921,7 @@ TiffRasterData getFloatingPointRasterData(
 
         final ImageDataReader dataReader = imageData.getDataReader(directory,
             photometricInterpreter, bitsPerPixel, bitsPerSample, predictor,
-            samplesPerPixel, width, height, compression, byteOrder);
+          samplesPerPixel, width, height, compression, 1, byteOrder);
```
kinow, line 924: Is `1` the equivalent of `TiffTagConstants.PLANAR_CONFIGURATION_VALUE_CHUNKY`? If so, I think we should use the constant?

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
```diff
@@ -237,27 +240,51 @@ private void interpretStrip(
...
-                    (int) bytesPerStrip, width, (int) rowsInThisStrip);
-
-            interpretStrip(
-                    imageBuilder,
-                    decompressed,
-                    (int) pixelsPerStrip,
-                    height);
-
+      throws ImageReadException, IOException {
+        if (planarConfiguration != 2) {
```
kinow, line 244: Is `1` the equivalent of `TiffTagConstants.PLANAR_CONFIGURATION_VALUE_PLANAR`? If so, I think we should use the constant?

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
```diff
@@ -291,25 +318,50 @@ public BufferedImage readImageData(final Rectangle subImage)
...
-
-            final byte[] decompressed = decompress(compressed, compression,
-                    (int) bytesPerStrip, width, (int) rowsInThisStrip);
-
-            interpretStrip(
-                    workingBuilder,
-                    decompressed,
-                    (int) pixelsPerStrip,
-                    yLimit);
+        if (planarConfiguration != 2) {
```
kinow, line 321: Is `1` the equivalent of `TiffTagConstants.PLANAR_CONFIGURATION_VALUE_PLANAR`? If so, I think we should use the constant?

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
```diff
@@ -291,25 +318,50 @@ public BufferedImage readImageData(final Rectangle subImage)
...
+                    final byte[] decompressed = decompress(compressed, compression,
+                      (int) bytesPerStrip, width, (int) rowsInThisStrip);
+                    int index = iPlane;
+                    for (int i = 0; i < decompressed.length; i++) {
+                        b[index] = decompressed[i];
+                        index += 3;
+                    }
+                }
+                interpretStrip(workingBuilder, b, (int) pixelsPerStrip, height);
+            }
```
kinow, line 364: This code here (`readImageData(final Rectangle subImage)`) looks a lot like the code above for `readImageData(final ImageBuilder imageBuilder)`. Should we make a function to read strips and re-use in both methods?

src/test/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStripsTest.java
```diff
@@ -24,7 +24,7 @@
     @Test
     public void testApplyPredictor() throws Exception {
         final int[] bitsPerPixel = {1,2,3};
-        final DataReaderStrips strips = new DataReaderStrips(null, null, 3, bitsPerPixel, 2, 4, 0, 3, 1, 1, null, 2, null);
+        final DataReaderStrips strips = new DataReaderStrips(null, null, 3, bitsPerPixel, 2, 4, 0, 3, 1, 1, 1, null, 2, null);
```
kinow, line 27: Can we use the constant `TiffTagConstants.PLANAR_CONFIGURATION_VALUE_CHUNKY` (not sure if the `1` here corresponds to that constant's value?)?

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
gwlucastrig: These are two methods have a lot of redundancy and it would be better to combine them into one.  Unfortunately, they are both public APIs, so I'm not sure we could change their external appearance.   Of course, we could create a single, consolidated method that does the processing. Then we could rewrite the existing methods to be "wrapper" methods that called the new consolidated method.

I think it might be better to treat that change as a separate pull request.

I contributed the "sub-image" logic a number of years ago. At the time, there was a certain amount of resistance to the idea of a sub-image API, but I thought it was necessary because may GeoTIFFs are huge (I'd been working with some that are 20000-by-20000 pixels large).  Anyway, I think I left the original method in place because the code was simpler than the sub-image code and I thought it would serve as a reference for anyone who was trying to figure out how it worked.  But I think you are right and the time has come to clean things up.

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageData.java
gwlucastrig: Some of the spacing got accidentally modified when I edited file.  I thought I cleaned them all, but I missed this one. Sorry.

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
gwlucastrig: You are absolutely right about using the TiffTagConstant values in the places you cited.  Actually, I wanted to make line 244 test for planarConfiguration == "Chunky" rather than != "Planar", but I was afraid that there might be places in the code where the planarConfiguration variable didn't get set correctly.  So I thought this would be safer.

If we ever get our hands on some additional test files, I might revisit this.  But that may be unlikely...  The Tiff Specification itself describes planar-configuration 2 as follows: _PlanarConfiguration=2 is not currently in widespread use and it is not recommended for general interchange. It is used an extension and Baseline TIFF readers are not required to support it._

src/test/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStripsTest.java
gwlucastrig: Agreed. Good catch.

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
gwlucastrig: Agreed. Good catch.

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
gwlucastrig: Agreed. Good catch.

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderStrips.java
kinow: >I think it might be better to treat that change as a separate pull request.

:+1:

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageData.java
```diff
@@ -53,14 +53,17 @@ public boolean stripsNotTiles() {
...
         public ImageDataReader getDataReader(final TiffDirectory directory,
-                final PhotometricInterpreter photometricInterpreter,
-                final int bitsPerPixel, final int[] bitsPerSample, final int predictor,
-                final int samplesPerPixel, final int width, final int height, final int compression,
-            final ByteOrder byteOrder) throws IOException, ImageReadException {
+          final PhotometricInterpreter photometricInterpreter,
+          final int bitsPerPixel, final int[] bitsPerSample, final int predictor,
+          final int samplesPerPixel, final int width, final int height,
+          final int compression,
+          final int planarConfiguration,
```
kinow, line 60: @gwlucastrig last change before merging this PR. We have some enum's like `DisposalMethod`, `IptcTypes`, png `FilterType`, `TiffDirectoryType`…

I think we should create some enum like `TiffPlanarConfiguration` for this setting. Do you think that would i) make sense and ii) work?

Nothing to add besides that, we have tests, code is in order, bug appears to be fixed.

gwlucastrig: Making planar configuration an enum would be a good idea.  I see that you self-assigned this.  Does that mean you'll be making the change?

In the future, making predictor and compression also be enum's would also make sense.  When I am reading the data-reader code I often lose track of what the numeric values indicate.

kinow: >Making planar configuration an enum would be a good idea. I see that you self-assigned this. Does that mean you'll be making the change?

No, sorry it's just an old habit. I believe I cannot assign the issue to you as you are an external contributor. I normally assign issues & PR's to myself so other common developers are aware I'm interested in working/finishing work on some issue.

gwlucastrig:
Okay.   I will update the pull request to use the TiffPlanarConfiguration
enumerated type.

gwlucastrig:
I believe this completes the work for Imaging-265.

kinow (approved):
Looks great! Merging. Thanks @gwlucastrig !

kinow:
Merged. Here's what I did to merge @gwlucastrig 

- checked out branch locally
- added a commit with the `src/changes/changes.xml` entry for this issue
- squashed commits
- reworded commit to match the change, using the [IMAGING-265] prefix

These are not a must-have, but really nice-to-have, as other components follow this pattern. This helps other common developers to maintain the code.

:+1: just because you are already familiar with most of the ASF development process, so I thought it would be interesting to share more of what happens with the PR's/issues with you :-) 

Thanks again for the contribution!
Bruno

gwlucastrig:
Thanks for the information on merge procedures.  The better I understand
the process, the easier it will be next time.

Gary

PR #106, opened 2020-10-30
[Issue-216] Support alpha channel in TIFF RGB formats

gwlucastrig:
Second attempt at a pull request in support of Issue-216

### DIFF
--- a/src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
+++ b/src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
@@ -40,8 +40,10 @@
  */
 package org.apache.commons.imaging.common;
 
+import java.awt.color.ColorSpace;
 import java.awt.image.BufferedImage;
 import java.awt.image.ColorModel;
+import java.awt.image.DataBuffer;
 import java.awt.image.DataBufferInt;
 import java.awt.image.DirectColorModel;
 import java.awt.image.Raster;
@@ -58,6 +60,7 @@ public class ImageBuilder {
     private final int width;
     private final int height;
     private final boolean hasAlpha;
+    private final boolean isAlphaPremultiplied;
 
     /**
      * Construct an ImageBuilder instance
@@ -68,6 +71,38 @@ public class ImageBuilder {
      * requirements for the ImageBuilder or resulting BufferedImage.
      */
     public ImageBuilder(final int width, final int height, final boolean hasAlpha) {
+        checkDimensions(width, height);
+
+        data = new int[width * height];
+        this.width = width;
+        this.height = height;
+        this.hasAlpha = hasAlpha;
+        this.isAlphaPremultiplied = false;
+    }
+
+
+    /**
+     * Construct an ImageBuilder instance
+     * @param width the width of the image to be built
+     * @param height the height of the image to be built
+     * @param hasAlpha indicates whether the image has an alpha channel
+     * (the selection of alpha channel does not change the memory
+     * requirements for the ImageBuilder or resulting BufferedImage.
+     * @param isAlphaPremultiplied indicates whether alpha values are
+     * pre-multiplied; this setting is relevant only if alpha is true.
+     *
+     */
+    public ImageBuilder(final int width, final int height,
+        final boolean hasAlpha, boolean isAlphaPremultiplied) {
+        checkDimensions(width, height);
+        data = new int[width * height];
+        this.width = width;
+        this.height = height;
+        this.hasAlpha = hasAlpha;
+        this.isAlphaPremultiplied = isAlphaPremultiplied;
+    }
+
+    private void checkDimensions(int width, int height) {
         if (width <= 0) {
             throw new RasterFormatException("zero or negative width value");
         }
@@ -75,10 +110,6 @@ public class ImageBuilder {
             throw new RasterFormatException("zero or negative height value");
         }
 
-        data = new int[width * height];
-        this.width = width;
-        this.height = height;
-        this.hasAlpha = hasAlpha;
     }
 
     /**
@@ -131,6 +162,39 @@ public class ImageBuilder {
         return makeBufferedImage(data, width, height, hasAlpha);
     }
 
+    /**
+     * Performs a check on the specified sub-region to verify
+     * that it is within the constraints of the ImageBuilder bounds.
+     *
+     * @param x the X coordinate of the upper-left corner of the
+     * specified rectangular region
+     * @param y the Y coordinate of the upper-left corner of the
+     * specified rectangular region
+     * @param w the width of the specified rectangular region
+     * @param h the height of the specified rectangular region
+     */
+    private void checkBounds(int x, int y, int w, int h) {
+        if (w <= 0) {
+            throw new RasterFormatException("negative or zero subimage width");
+        }
+        if (h <= 0) {
+            throw new RasterFormatException("negative or zero subimage height");
+        }
+        if (x < 0 || x >= width) {
+            throw new RasterFormatException("subimage x is outside raster");
+        }
+        if (x + w > width) {
+            throw new RasterFormatException(
+                "subimage (x+width) is outside raster");
+        }
+        if (y < 0 || y >= height) {
+            throw new RasterFormatException("subimage y is outside raster");
+        }
+        if (y + h > height) {
+            throw new RasterFormatException(
+                "subimage (y+height) is outside raster");
+        }
+    }
 
      /**
      * Gets a subset of the ImageBuilder content using the specified parameters
@@ -150,28 +214,8 @@ public class ImageBuilder {
      *         within this ImageBuilder
      */
     public ImageBuilder getSubset(final int x, final int y, final int w, final int h) {
-        if (w <= 0) {
-            throw new RasterFormatException("negative or zero subimage width");
-        }
-        if (h <= 0) {
-            throw new RasterFormatException("negative or zero subimage height");
-        }
-        if (x < 0 || x >= width) {
-            throw new RasterFormatException("subimage x is outside raster");
-        }
-        if (x + w > width) {
-            throw new RasterFormatException(
-                    "subimage (x+width) is outside raster");
-        }
-        if (y < 0 || y >= height) {
-            throw new RasterFormatException("subimage y is outside raster");
-        }
-        if (y + h > height) {
-            throw new RasterFormatException(
-                    "subimage (y+height) is outside raster");
-        }
-
-        ImageBuilder b = new ImageBuilder(w, h, hasAlpha);
+        checkBounds(x, y, w, h);
+        ImageBuilder b = new ImageBuilder(w, h, hasAlpha, isAlphaPremultiplied);
         for(int i=0; i<h; i++){
             int srcDex = (i+y)*width+x;
             int outDex = i*w;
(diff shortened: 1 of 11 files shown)


src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
```diff
@@ -68,17 +71,45 @@
      * requirements for the ImageBuilder or resulting BufferedImage.
      */
     public ImageBuilder(final int width, final int height, final boolean hasAlpha) {
+        checkDimensions(width, height);
```
kinow, line 74: Great idea! Thanks!!! :+1:

src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
```diff
@@ -242,16 +266,29 @@ private BufferedImage makeBufferedImage(
         WritableRaster raster;
         final DataBufferInt buffer = new DataBufferInt(argb, w * h);
         if (useAlpha) {
-            colorModel = new DirectColorModel(32, 0x00ff0000, 0x0000ff00,
-                    0x000000ff, 0xff000000);
-            raster = Raster.createPackedRaster(buffer, w, h,
-                    w, new int[]{0x00ff0000, 0x0000ff00, 0x000000ff,
-                            0xff000000}, null);
+            colorModel = new DirectColorModel(
+                    ColorSpace.getInstance(ColorSpace.CS_sRGB),
```
kinow, line 270: Why is it sRGB? I couldn't see in the diff if we had it before. But why couldn't it be just `TYPE_RGB` or some other type?

src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java
```diff
@@ -287,7 +330,20 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
         final int width = src.getWidth();
         final int height = src.getHeight();
 
-        int compression = TIFF_COMPRESSION_LZW; // LZW is default
+        final ColorModel cModel = src.getColorModel();
+        final boolean hasAlpha = cModel.hasAlpha() && checkForActualAlpha(src);
```
kinow, line 334: Initially I thought maybe we didn't have to check for actual alpha in the image, since the colour model was already defining that it has alpha.

But this is for writing images, so it should be used just once... @gwlucastrig worst case scenario I guess would be a super large file, with no alpha. In that case We end up reading the whole image in memory (in chunks, but nevertheless...).

Do you think this could be a problem in the future? I'm fine with this change for now. If anybody reports any slowness we can just use a parameter to control whether this check is performed or not :man_shrugging: WDYT?

kinow (approved):
Looks good to me! A few minor comments, happy to merge once @gwlucastrig has replied to my comments :tada: 

Thanks Gary!
Bruno

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderTiled.java
```diff
@@ -141,32 +141,37 @@ private void interpretTile(final ImageBuilder imageBuilder, final byte[] bytes,
...
+                for (int i = i0; i < i1; i++) {
+                    int k = (i - i0) * tileWidth * 4;
+                    for (int j = j0; j < j1; j++, k += 4) {
+                        final int rgb
+                            = ((bytes[k] & 0xff) << 16)
+                            | ((bytes[k + 1] & 0xff) << 8)
+                            | (bytes[k + 2] & 0xff)
+                            | (bytes[k + 3] << 24);
+                        imageBuilder.setRGB(j, i, rgb);
+                    }
```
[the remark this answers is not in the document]
gwlucastrig: Data reader strips and data reader tiles have always had a lot of overlap.  Really, a strip and a tile are both rectangles filled with pixels... Though TIFF has slightly different rules for cases that run off the edges of images...  Someday, I'd like to take a look at just consolidating the two classes. But it's going to be a pretty big change to the existing reader classes.

src/main/java/org/apache/commons/imaging/formats/tiff/constants/TiffTagConstants.java
```diff
@@ -347,6 +347,8 @@
     public static final TagInfoShorts TIFF_TAG_EXTRA_SAMPLES = new TagInfoShorts(
             "ExtraSamples", 0x152, -1,
             TiffDirectoryType.TIFF_DIRECTORY_ROOT);
+    public static final int EXTRA_SAMPLE_ASSOCIATED_ALPHA = 1;
+    public static final int EXTRA_SAMPLE_UNASSOCIATED_ALPHA = 2;
```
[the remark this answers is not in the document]
gwlucastrig: Associated and Unassociated are the names they use in the TIFF specification.  In playing with the Java BufferedImage class, there seems to be a bias toward unassociated alpha (non-pre-multiplied).  For example, at least in the OpenJDK Java I use, if you have a pre-multiplied images and call getRGB on it, the pixel values that come back are adjusted to look like non-pre-multiplied.  

Not sure I answered your question here....  Sorry.

src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
gwlucastrig: I think that sRGB is the default for Java and would give us the best performance.  I can double check this.

src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java
gwlucastrig: The main reason I checked for actual alpha is that I'd seen a bunch of images of TYPE_INT_ARGB that had all the alpha values set to 0xff anyway.  I thought it would help reduce the size of the output file to leave it out if the image didn't actually use transparency.

Also, I've seen a bug in java's ImageIO class where if you try to store a BufferedImage of TYPE_INT_ARGB to a JPEG file, Java messes up the color.   That bug has been there like forever...  So I thought that it might avoid bugs for applications if by storing our TIFFs so that when they got read into BufferedImages by an application they would end up just simple TYPE_INT_RGB.  That way, an application that used Commons Imaging to read a TIFF and then ImageIO to write a JPEG might have better luck.

src/main/java/org/apache/commons/imaging/formats/tiff/constants/TiffTagConstants.java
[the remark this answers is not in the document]
kinow: :+1: looks like our constant will be similar to what others will find in OpenJDK, so all good. Thanks Gary!

src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
kinow: I normally default to RGB, but for no special reason. I think sRGB is some standard related to RGB, but I can't really remember ever setting a file to be saved as sRGB :thinking: so asking more out of curiosity. If this needs changing later, it wouldn't break binary backward compatibility, so we can fix it even after 1.0 release (if necessary)

src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java
gwlucastrig: Also, the ImageWriterBase always reads the whole image into memory.  That's one of my motivations for coming up with a new generation of image writer class in the JIRA item I posted recently.  The checkForActualAlpha has logic so that it reads a huge image in parts, so that limits the memory a little. Still, you are correct that it does mean two passes through the image data. But I think it is probably worth it to produce a smaller image on output.

kinow: :+1: thanks for the detailed reply @gwlucastrig ! Is there some of what you said, that could go into a comment in this new code? For any new developers (or old ones with a horrible memory like me) of Imaging to be aware of why it's working this way.

src/main/java/org/apache/commons/imaging/common/ImageBuilder.java
gwlucastrig: Just finished looking through Java documentation and I got to say I cannot believe how complicated this stuff is. Who would of thought something like color would have such an elaborate set of support classes.

Anyway, I'm not really sure of the implications of this, but the Java API documentation for the Color class says "The default color space for the Java 2D(tm) API is sRGB, a proposed standard RGB color space. For further information on sRGB, see http://www.w3.org/pub/WWW/Graphics/Color/sRGB.html ".   That's what led me to use the declaration you saw.  I also ran a quick test program to see what was the default used by BufferedImage:

       BufferedImage bImage = new BufferedImage(10, 10, BufferedImage.TYPE_INT_RGB);
        ColorSpace cSpace = bImage.getColorModel().getColorSpace();
        ColorSpace linearRGB = ColorSpace.getInstance(ColorSpace.CS_LINEAR_RGB);
        ColorSpace sRGB = ColorSpace.getInstance(ColorSpace.CS_sRGB);
        System.out.println("Compare to Linear RGB " + cSpace.equals(linearRGB));  // printed out "false"
        System.out.println("Compare to sRGB       " + cSpace.equals(sRGB));       // printed out "true"

kinow: If a company like Microsoft, Google, Facebook, etc, invested some money and time in organising all the image standards into a single, uniform website, like JavaScript MDN, W3C specs, etc, that would make writing libraries like Common Imaging much easier (and more sane :laughing: ).

Thanks for looking into this. Looks like we are good staying up close with what the JDK does, unless a user has a valid case for us using another type.

src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java
gwlucastrig: I've got some comments that I will incorporate into the code in my next PR.  Unfortunately, once again I've broken git and I can't get it to accept more changes to the current PR.  Clearly, there is something systematic I'm doing wrong and I am trying to figure out what that is.

In the meantime...  Here's the text I've written.  It would go into TiffImageWriterBase.java right before the call to checklForActualAlpha.

       // If the source image has a color model that supports alpha,
        // this module performs a call to checkForActualAlpha() to see whether
        // the image that was supplied to the API actually contains
        // non-opaque data in its alpha channel.  It is common for applications
        // to create a BufferedImage using TYPE_INT_ARGB, and fill the entire
        // image with opaque pixels. In such a case, the file size of the output
        // can be reduced by 25 percent by storing the image in an 3-byte RGB
        // format. This approach will also make a small reduction in the runtime
        // to read the resulting file when it is accessed by an application.

kinow: @gwlucastrig I just had a look at your forked repo, and I believe I am able to push changes to your branch Gary. That's because by default pull requests allow contributors of a repository to push against the original branch of the pull request.

In that case, you should be able to push to your branch too. In a brand new directory, here's what I'd do:

```bash
git clone git@github.com:gwlucastrig/commons-imaging.git -o origin
git checkout issue-216
vim src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java 
git add .
git commit -m '[IMAGING-216] Updated documentation for TIFF writer base class'
git push -u origin issue-216
```

After that the new commit would appear here.

I used a new directory as that's what I'd have to do in my machine to push to your branch. But if you have a working directory with commons-imaging, you can try to use that to experiment with git and try to push to this PR.

gwlucastrig:
Okay.  I tried that and it seems to have worked. I hope that will finish
off this particular issue and I will be on to the next one...

src/main/java/org/apache/commons/imaging/formats/tiff/write/TiffImageWriterBase.java
```diff
@@ -330,9 +330,19 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
...
 
+        // If the source image has a color model that supports alpha,
+        // this module performs a call to checkForActualAlpha() to see whether
+        // the image that was supplied to the API actually contains
+        // non-opaque data in its alpha channel. It is common for applications
+        // to create a BufferedImage using TYPE_INT_ARGB, and fill the entire
+        // image with opaque pixels. In such a case, the file size of the output
+        // can be reduced by 25 percent by storing the image in an 3-byte RGB
+        // format. This approach will also make a small reduction in the runtime
+        // to read the resulting file when it is accessed by an application.
```
kinow, line 341: 🎉

kinow:
Added changelog, updated JIRA, and merged. Thanks @gwlucastrig !

PR #108, opened 2020-11-22
Issue-266 access full precision of TIFF samples

gwlucastrig:
This change will allow applications to read elevation data from NASA's Shuttle Radar Topography Mission (SRTM).  This change will give Commons Imaging users access to more than ten thousand high-resolution elevation data products covering the entire Earth,

The key changes are the addition of a new method as part of TiffDirectory which returns an instance of the ImageBuilder class.  The ImageBuilder can be accessed to read elevation samples from the 2-byte raster data.  I also had to enhance the internal data-reader operations to preserve the full precision of the source data (the current version reduces all samples to one byte).  I have included a Junit test and a sample file called Sample16BitIntStrips.tiff.  I am still working on a sample file organized in Tiles, so the changes to the DataReaderTiles class can be exercised properly (there will be a small drop off in coverage until I get the new sample file ready).

I also plan on submitting an example application showing how to use the API.


src/main/java/org/apache/commons/imaging/formats/tiff/TiffDirectory.java
```diff
@@ -914,4 +915,41 @@ public boolean hasTiffFloatingPointRasterData() throws ImageReadException {
...
+     * @throws ImageReadException in the event of incompatible or malformed data
+     * @throws IOException in the event of an I/O error
+     */
+    public ImageBuilder getImageBuilder(
+        final Map<String, Object> params)
+        throws ImageReadException, IOException {
+        TiffImageParser parser = new TiffImageParser();
+        return parser.getImageBuilder(this, headerByteOrder, params);
+    }
+
```
kinow, line 954: @gwlucastrig maybe we could rethink this? It strikes me as strange/risky having the `ImageBuilder` and `TiffDirectory` so tightly tied. I think the builder creates/uses directories; and—if that relationship really exists—then it would be a circular dependency by accessing the Builder from the Directory. Will take a better look later with the IDE :+1:

src/main/java/org/apache/commons/imaging/formats/tiff/photometricinterpreters/PhotometricInterpreterTranscribeSample0.java
```diff
@@ -0,0 +1,111 @@
...
+ * <p>
+ * Note that this implementation only processes the first sample for a pixel.
+ * Developers requiring access to additional samples may use this class as
+ * an example.  However, the ImageBuilder class currently has the limitation
+ * that it cannot store more than one 32 bit integer per pixel or
+ * raster cell.
+ * <p>
+ * The data stored by this class is treated as a signed integer value.
+ */
+public class PhotometricInterpreterTranscribeSample0 extends PhotometricInterpreter {
```
kinow, line 42: Maybe we could have a different name? Instead of `PhotometricInterpreterTranscribeSample0`, would it make sense to have something like `...Basic`, `...Default`, etc? This way we avoid having future classes with names that do not match the pattern `...0`, `...1`, etc.

kinow (commented):
Had a cursory look @gwlucastrig , looks good, and great improvement for Imaging! :man_astronaut: :star: 

Few comments, but will use the IDE to confirm what I have said :+1:

gwlucastrig:
A better name for the transcriber class would be a good idea...  in terms
of the directories, I have no strong preference.  One thing to keep in mind
is the a TIFF can be a bundle of multiple images and the directory (the
IFD) is the main unit of organization.  But the nesting of classes in
Commons Imaging is a bit awkward and could benefit from some TLC.

src/main/java/org/apache/commons/imaging/formats/tiff/photometricinterpreters/PhotometricInterpreterTranscribeSample0.java
gwlucastrig: I got rid of the zero at the end of the name.  The class name looks a lot cleaner without it. Good suggestion!

gwlucastrig:
I added some JUnit tests.  I still want to write an example application showing how to use the API to get data.  I've been using it myself quite a bit to read elevation data for testing my data compression project (the Auckland sample compresses down to 1.2539 bits per elevation). But I need to write a generic example that can be run with just Commons Imaging.

I have been thinking about your suggestion regarding the directories.  I am willing to modify this, but I'm not sure what would be better.  As I said, the main issue is that in a TIFF file, the directory is the thing that bundles the image and the metadata (the TIFF spec calls them "Image File Directories", or IFDs).   You can see examples of how they are used in the example classes in commons-imaging\src\test\java\org\apache\commons\imaging\examples\tiff.  ReadTagsAndImages.java loops through the directories in a file getting metadata tags and imagery from each.  ExampleReadFloatingPointData.java reads the floating-point raster data from the TIFF directory.  In fact, the original U.S. Geological Survey products that got me started working on the floating-point format actually contain 6 separate directories, each at a different level of resolution.

PR #112, opened 2020-12-13
[IMAGING-273] - Minor improvement

arturobernalg:
Minor improvement:

- extra computations by getting the string size only once.
- remove unused imports
- add final String
- add final method and var
- fix javaDoc

https://issues.apache.org/jira/browse/IMAGING-273

### DIFF
--- a/src/main/java/org/apache/commons/imaging/common/itu_t4/T4AndT6Compression.java
+++ b/src/main/java/org/apache/commons/imaging/common/itu_t4/T4AndT6Compression.java
@@ -139,8 +139,8 @@ public final class T4AndT6Compression {
     public static byte[] decompressModifiedHuffman(final byte[] compressed,
             final int width, final int height) throws ImageReadException {
         try (ByteArrayInputStream baos = new ByteArrayInputStream(compressed);
-                BitInputStreamFlexible inputStream = new BitInputStreamFlexible(baos);
-                BitArrayOutputStream outputStream = new BitArrayOutputStream()) {
+             BitInputStreamFlexible inputStream = new BitInputStreamFlexible(baos);
+             BitArrayOutputStream outputStream = new BitArrayOutputStream()) {
             for (int y = 0; y < height; y++) {
                 int color = WHITE;
                 int rowLength;
@@ -246,7 +246,7 @@ public final class T4AndT6Compression {
     private static int compressT(final int a0, final int a1, final int b1, final BitArrayOutputStream outputStream,final  int codingA0Color, final int[] codingLine ){
           final int a1b1 = a1 - b1;
           if (-3 <= a1b1 && a1b1 <= 3) {
-              T4_T6_Tables.Entry entry;
+              final T4_T6_Tables.Entry entry;
               switch (a1b1) {
             case -3:
                 entry = T4_T6_Tables.VL3;
@@ -394,8 +394,8 @@ public final class T4AndT6Compression {
                         int b1 = nextChangingElement(referenceLine, referenceA0Color, 0);
                         int b2 = nextChangingElement(referenceLine, 1 - referenceA0Color, b1 + 1);
                         for (int a0 = 0; a0 < width;) {
-                            int a1;
-                            int a2;
+                            final int a1;
+                            final int a2;
                             entry = CONTROL_CODES.decode(inputStream);
                             if (entry == T4_T6_Tables.P) {
                                 fillRange(outputStream, referenceLine, a0, b2, codingA0Color);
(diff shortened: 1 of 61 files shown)


kinow (commented):
Hello @arturobernalg !

Thanks for your pull request!

I had a quick cursory look—it's late Sunday evening, so my eyes are not that sharp for a proper review.

There are some improvements that we could fix right away. Like the extra spaces, or Javadocs.

For `.length` and `final`, do you think that's really necessary @arturobernalg ? For class fields I understand `final` being used to enforce immutable objects… but some variables appear to have a method-scope.

In those cases, I think `final` isn't effectively used, nor does it help the JVM.

The `.length` is also already immutable in the array, and I **think** the JVM optimizes its access (although I haven't looked at the generated bytecode).

Changes with multiple lines like this have pros and cons. The main issue for us, is i) possible conflicts in pull requests, which maintainers have to ask contributors to rebase and fix conflicts, or do it themselves; and ii) having to go through past changes when doing some major refactoring, seeing what's in each commit.

That's why we try to have the smallest impact in each pull request :slightly_smiling_face: so that we can merge the contributed code, and also be confident it won't cause any issues in the future, no matter the size of the change :+1: 

Thanks!
Bruno

p.s.: last week before holidays, where I will have a lot more of spare time for OSS, especially [imaging] :tada: so a proper review coming soon

arturobernalg:
Hi @kinow 
The idea of declaring variables final within methods is to prevent them from being modified within the scope by mistake. In this way, we force everyone who wants to work on the method not to modify it.

Pls let me know if you prefer other PR only with the change of the javadoc while you complete the review. 

I'm glad to know that your holidays are coming soon :)

src/main/java/org/apache/commons/imaging/common/itu_t4/T4AndT6Compression.java
```diff
@@ -139,8 +139,8 @@ private static void compress1DLine(final BitInputStreamFlexible inputStream,
     public static byte[] decompressModifiedHuffman(final byte[] compressed,
             final int width, final int height) throws ImageReadException {
         try (ByteArrayInputStream baos = new ByteArrayInputStream(compressed);
-                BitInputStreamFlexible inputStream = new BitInputStreamFlexible(baos);
-                BitArrayOutputStream outputStream = new BitArrayOutputStream()) {
```
kinow, line 143: :+1:

src/main/java/org/apache/commons/imaging/formats/jpeg/iptc/IptcParser.java
```diff
@@ -59,8 +59,8 @@
      * Block types (or Image Resource IDs) that are not recommended to be
      * interpreted when libraries process Photoshop IPTC metadata.
      *
-     * @see https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/
-     * @see https://issues.apache.org/jira/browse/IMAGING-246
+     * @see <a href="https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/"> Adobe Photoshop File Formats Specification</a>
+     * @see <a href="https://issues.apache.org/jira/browse/IMAGING-246>IMAGING-246</a>
```
kinow, line 63: Merged in your other PR :+1: (thanks!)

kinow (changes requested):
Hi @arturobernalg !

There is one change here that I think we could merge, that is about a formatting issue (from what I could see in the GitHub UI, looks like we have extra spaces, good catch!)

However, I am not really keen to add the remaining changes.

The access to `.length` shouldn't give any performance improvements. And it would be really likely to have a contributor to send a pull request with `variable.length` again. Trying to force users to use `int variableLength = variable.length;` isn't something that can be easily enforced.

I am not the only developer reviewing things. It is not something easy to have a linter or static analyzer to remind us of the way we want to use. So on and so on.

About the `final` modifier, it also wouldn't improve performance. It would enforce local variables to remain immutable. But it goes along what I said before, about enforcing it.

That would become a project preference. But it would be hard to make sure we always had `final` everywhere. And it could lead to unnecessary discussions here and mailing list.

I am +1 to using `final` if it will enforce immutability to the API users. e.g. when we enforce an object to be instantiated with values via constructor, providing only `getters`.

We have checkstyle enforcing some code style, and practices. And we keep an eye on things like thread-safety, performance issues, following rules like JIRA issues and changelog for some pull requests. Adding more things like remembering to look for `variableLength` and the `final` modifier would need other developers to be OK with this, and intend to keep enforcing during code reviews.

I hope that makes sense. If you want, you can leave this PR open. As I mentioned earlier, other developers work on this project too. So perhaps others would have a different opinion about it, and want to discuss this idea.

Thanks!
Bruno

arturobernalg:
Hi @kinow . 
I'm closing the PR. make sense what are you saying

PR #137, opened 2021-04-22
IMAGING-292 - Simplify Bitwise operation

arturobernalg:


aherbert:
The original code is clearer as to the intention. The compiler will get rid of these anyway and it will have no effect on the runtime.

arturobernalg:
it's clearer result[offset + 0]  than   result[offset] .... not agree..

aherbert:
It is more about the visual layout of either side of the operation:
```java
result[offset + 3] = (byte) (value >> 24);
result[offset + 2] = (byte) (value >> 16);
result[offset + 1] = (byte) (value >> 8);
result[offset + 0] = (byte) (value >> 0);
```
becomes:
```java
result[offset + 3] = (byte) (value >> 24);
result[offset + 2] = (byte) (value >> 16);
result[offset + 1] = (byte) (value >> 8);
result[offset    ] = (byte) (value);
```

If you keep the formatting using spaces for layout purposes where you have removed redundant operations that makes the code clearer.

If you are going to remove all the `+ 0` occurrences then you should remove all the ` >> 0` and ` << 0` bit shifts which are also redundant.

I believe this:
```java
                        final int index = (u << (precision * 2))
                                | (j << (precision * 1))
                                | (k << (precision * 0));
```
has more information about the code intension than:
```java
                        final int index = (u << (precision * 2))
                                | (j << precision)
                                | k;
```

If you prefer the later then please update the PR to remove all the shifts by zero and maintain alignment of the source for clarity.

arturobernalg:
> It is more about the visual layout of either side of the operation:
> 
> ```java
> result[offset + 3] = (byte) (value >> 24);
> result[offset + 2] = (byte) (value >> 16);
> result[offset + 1] = (byte) (value >> 8);
> result[offset + 0] = (byte) (value >> 0);
> ```
> 
> becomes:
> 
> ```java
> result[offset + 3] = (byte) (value >> 24);
> result[offset + 2] = (byte) (value >> 16);
> result[offset + 1] = (byte) (value >> 8);
> result[offset    ] = (byte) (value);
> ```
> 
> If you keep the formatting using spaces for layout purposes where you have removed redundant operations that makes the code clearer.
> 
> If you are going to remove all the `+ 0` occurrences then you should remove all the ` >> 0` and ` << 0` bit shifts which are also redundant.
> 
> I believe this:
> 
> ```java
>                         final int index = (u << (precision * 2))
>                                 | (j << (precision * 1))
>                                 | (k << (precision * 0));
> ```
> 
> has more information about the code intension than:
> 
> ```java
>                         final int index = (u << (precision * 2))
>                                 | (j << precision)
>                                 | k;
> ```
> 
> If you prefer the later then please update the PR to remove all the shifts by zero and maintain alignment of the source for clarity.

agree --> https://github.com/apache/commons-imaging/pull/138

PR #142, opened 2021-04-26
IMAGING-296 - Remove Redundant Collection operation

arturobernalg:


kinow (changes requested):
Not sure if I'll review/merge this one @arturobernalg . Several changes here are in the parameters. These will be removed in #116 . I'd have to handle the rebase/conflicts, so I think I'll wait until I have finished that PR before looking at this one again.

arturobernalg:
> Not sure if I'll review/merge this one @arturobernalg . Several changes here are in the parameters. These will be removed in #116 . I'd have to handle the rebase/conflicts, so I think I'll wait until I have finished that PR before looking at this one again.

HI @kinow Fully agree. We wait.

kinow (commented):
@arturobernalg turns out all of these operations are removed in #116 

That's because these operations are removing given parameters, but after #116 & IMAGING-159 we won't have a `params` hashmap any longer, but a Java POJO that won't need a set/unset :+1: 

Closing.

Thanks!

PR #154, opened 2021-06-15
Simplify conditions and avoid extra checks.

arturobernalg:


src/main/java/org/apache/commons/imaging/palette/LongestAxisMedianCut.java
```diff
@@ -43,7 +43,7 @@ public boolean performNextMedianCut(final List<ColorGroup> colorGroups, final bo
                 && colorGroup.alphaDiff > colorGroup.redDiff
                 && colorGroup.alphaDiff > colorGroup.greenDiff
                 && colorGroup.alphaDiff > colorGroup.blueDiff) {
-            doCut(colorGroup, ColorComponent.ALPHA, colorGroups, ignoreAlpha);
+            doCut(colorGroup, ColorComponent.ALPHA, colorGroups, false);
```
kinow, line 46: I think it was harmless to keep using `ignoreAlpha` here. There are other conditions in this if statement, and they are all using `ignoreAlpha`. Having that variable in all the conditions might make it easier to remove/update it later if necessary.

src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
```diff
@@ -447,7 +447,7 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
             final PaletteFactory paletteFactory = new PaletteFactory();
 
             if (hasAlpha) {
-                palette = paletteFactory.makeQuantizedRgbaPalette(src, hasAlpha, maxColors);
+                palette = paletteFactory.makeQuantizedRgbaPalette(src, true, maxColors);
```
kinow, line 450: I think it's fine to pass `hasAlpha` here. Easier to see (or guess) that that's a flag for transparency, especially in case we add more booleans to that method later.

kinow (approved):
Few comments, but everything else looks good. Thanks for the contributions @arturobernalg !

src/main/java/org/apache/commons/imaging/palette/LongestAxisMedianCut.java
arturobernalg: HI @kinow 
Changed
TY

src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
arturobernalg: HI @kinow 
Changed
TY

kinow (approved):

kinow:
Thanks a lot for fixing it so quickly, and also for squashing commits @arturobernalg :+1: merged!

PR #155, opened 2021-06-15
Remove redundant local variable.

arturobernalg:


src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -1003,7 +1003,6 @@ public void writeImage(final BufferedImage src, final OutputStream os, Map<Strin
                 // boolean LocalColorTableFlag = false;
                 final boolean interlaceFlag = false;
                 final boolean sortFlag = false;
-                final int sizeOfLocalColorTable = colorTableScaleLessOne;
```
kinow, line 1006: I think the author here intentionally made a redundant variable so other developers would understand the code more easily. Imaging has a lot of code with many variables, or methods implementing papers and using the variables as they appear in the papers.

So any help like this helps, so IMO we should keep issues like that when found by static analyzers / linters.

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/Dct.java
```diff
@@ -282,15 +282,14 @@ public static void inverseDCT8(final float[] vector) {
...
 
         // A2
         final float m3 = n1 + n2;
         final float m4 = n3 + a3;
         final float m5 = n1 - n2;
         final float m6 = n3 - a3;
-        // float m7 = n5 - n0;
-        final float neg_m7 = neg_n5 + n0;
+        // float m7 = b4 - n0;
+        final float neg_m7 = neg_b4 + n0;
```
kinow, line 292: Quite sure these variables were copied from a paper or specification document. I would prefer to leave as in the document (if that's the case), but will have to take a better look later to verify if that's really the case.

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
arturobernalg: HI @kinow 
Changed
TY

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/Dct.java
arturobernalg: HI @kinow 
Changed
TY

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
```diff
@@ -242,43 +242,41 @@ public boolean visitSegment(final int marker, final byte[] markerBytes,
         } else if (marker == JpegConstants.DQT_MARKER) {
             final DqtSegment dqtSegment = new DqtSegment(marker, segmentData);
             for (final QuantizationTable element : dqtSegment.quantizationTables) {
-                final DqtSegment.QuantizationTable table = element;
```
kinow, line 245: Or we could rename `element` to `table`? WDYT @arturobernalg ? This way the "Invalid quantization table identifier" message would clearly reflect that the identifier is from `table` (as in the variable name too).

Table is also a common term in many image processing documents (I guess that's why the author of the code created the alias-variable).

src/main/java/org/apache/commons/imaging/formats/png/PngImageParser.java
```diff
@@ -254,9 +254,8 @@ public void readSignature(final InputStream is) throws ImageReadException,
         }
 
         final PngChunkIccp pngChunkiCCP = (PngChunkIccp) chunks.get(0);
-        final byte[] bytes = pngChunkiCCP.getUncompressedProfile(); // TODO should this be a clone?
 
-        return (bytes);
+        return (pngChunkiCCP.getUncompressedProfile());
```
kinow, line 258: This removes the `TODO` comment that was left there by Sebb during one of his reviews. Better leave it there until someone else has time to ponder on that item, and maybe remove the redundant variable.

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
```diff
@@ -376,16 +374,14 @@ private void readMCU(final JpegInputStream is, final int[] preds, final Block[]
                                 is,
                                 huffmanACTables[scanComponent.acCodingTableSelector]);
                         final int ssss = rs & 0xf;
-                        final int rrrr = rs >> 4;
-                        final int r = rrrr;
```
kinow, line 380: It does look odd, but if you look at the comments near this code, they are pointing to which page of a document the author was referencing when writing it. These variable names, though obscure or redundant, are more helpful for devs familiar with that document.

See this Python examples: 

- https://github.com/Exaphis/python-jpeg-decoder/blob/849c3d0bb32355e44a40364ea0498e6f59443716/jpeg.py#L302
- https://github.com/yohhoy/picojdec/blob/316ef275c757bb67d4398980660f42915f67901c/picojdec.py#L152

Or these excerpts from a specification document:

image

image

I use Eclipse + IntelliJ (and other tools like CPD, SpotBugs, stan4j, sonarqube, etc) but for [imaging] I ignore several issues to avoid reducing the clarity of the code for future developers familiar with image processing :+1: 

So sorry for turning down changes like this @arturobernalg , even though you are correct some variables are redundant.

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderTiled.java
```diff
@@ -85,26 +85,24 @@ private void interpretTile(final ImageBuilder imageBuilder, final byte[] bytes,
         if (sampleFormat == TiffTagConstants.SAMPLE_FORMAT_VALUE_IEEE_FLOATING_POINT) {
             // tileLength: number of rows in tile
             // tileWidth:  number of columns in tile
-            final int i0 = startY;
```
kinow, line 88: Didn't search any specification document, but I would probably leave these variables as-is as they appear to be the type of vars copied from some existing code or document...

src/main/java/org/apache/commons/imaging/formats/rgbe/RgbeImageParser.java
```diff
@@ -102,15 +102,14 @@ public BufferedImage getBufferedImage(final ByteSource byteSource, final Map<Str
...
-            final BufferedImage ret = new BufferedImage(new ComponentColorModel(
+            return new BufferedImage(new ComponentColorModel(
                     ColorSpace.getInstance(ColorSpace.CS_sRGB), false, false,
                     Transparency.OPAQUE, buffer.getDataType()),
                     Raster.createWritableRaster(
                             new BandedSampleModel(buffer.getDataType(),
                                     info.getWidth(), info.getHeight(), 3),
                             buffer,
                             new Point()), false, null);
-            return ret;
```
kinow, line 113: :+1:

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
arturobernalg: HI @kinow 
I am agree with `table` name. Seem more logical.

src/main/java/org/apache/commons/imaging/formats/tiff/datareaders/DataReaderTiled.java
arturobernalg: Done

src/main/java/org/apache/commons/imaging/formats/png/PngImageParser.java
arturobernalg: Reverted
TY

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
```diff
@@ -262,23 +261,22 @@ public boolean visitSegment(final int marker, final byte[] markerBytes,
         } else if (marker == JpegConstants.DHT_MARKER) {
             final DhtSegment dhtSegment = new DhtSegment(marker, segmentData);
             for (final HuffmanTable element : dhtSegment.huffmanTables) {
```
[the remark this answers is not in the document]
kinow: Sorry, came to review this one today, and realized I had an old comment pending to be submitted.

PR #161, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
According to the official document [Fast Finishing](https://docs.travis-ci.com/user/build-matrix/#fast-finishing), if some rows in the build matrix are allowed to fail, we can add fast_finish: true to the .travis.yml to get faster feedbacks.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ jdk:
   - openjdk-ea
 
 matrix:
+  fast_finish: true
   allow_failures:
     - jdk: openjdk-ea
 


kinow (commented):
While I find this option useful, for Imaging I think I'd prefer to see whether the other JVM settings will fail too. Since we have just one dimension in the matrix (the jvm) and not a combination (e.g. browser version, maven version, etc) I think it's OK to leave the fast_finish: false (default) for now.

PR #162, opened 2021-08-17
IMAGING-308 Fix to unit test for windows

kinow:


kinow:
@gwlucastrig I went to your repository, and realized it's forked from my fork. We need to fix that, but can do that later so you don't lose anything.

So I went to pull requests in your repo, and clicked on create new pull request. By default it shows the dropbox with my fork as base. I just changed that to apache/commons-imaging, and created this PR 👍

gwlucastrig:
>>>  I went to your repository, and realized it's forked from my fork

Yes.   I figured that was the wrong way to do things, but it was the only
way I could get github to work.

When I tried forking from the main repository, I couldn't get github to
accept pull requests to your branch.  I would think that the idea that a
developer would need to contribute to two development paths would be
commonplace. So clearly I missed something.  Though Google and Github help
did not reveal what that might be.

kinow:
Ah, I see. In that case, it's better that you fork apache/commons-imaging (later after we merge your PR's if you'd like to wait), then create PR's against my fork if necessary the same way I created to a different base repo.

PR #163, opened 2021-08-17
IMAGING-310 JpegImageParser.getImageInfo() Refactor colorType logic to first look at numberOfComponents

jephillips34:
fix for IMAGING-310 JpegImageParser: Grayscale JPEG file with app14Segment returns ColorType.UNKNOWN

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
@@ -824,44 +824,44 @@ public class JpegImageParser extends ImageParser implements XmpEmbeddable {
 
         // See http://docs.oracle.com/javase/6/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html#color
         ImageInfo.ColorType colorType = ImageInfo.ColorType.UNKNOWN;
-        // Some images have both JFIF/APP0 and APP14.
-        // JFIF is meant to win but in them APP14 is clearly right, so make it win.
-        if (app14Segment != null && app14Segment.isAdobeJpegSegment()) {
-            final int colorTransform = app14Segment.getAdobeColorTransform();
-            switch (colorTransform) {
-            case App14Segment.ADOBE_COLOR_TRANSFORM_UNKNOWN:
-                if (numberOfComponents == 3) {
-                    colorType = ImageInfo.ColorType.RGB;
-                } else if (numberOfComponents == 4) {
-                    colorType = ImageInfo.ColorType.CMYK;
+        switch (numberOfComponents) {
+        case 1:
+            colorType = ImageInfo.ColorType.GRAYSCALE;
+            break;
+        case 2:
+            colorType = ImageInfo.ColorType.GRAYSCALE;
+            transparent = true;
+            break;
+        case 3:
+        case 4:
+            // Some images have both JFIF/APP0 and APP14.
+            // JFIF is meant to win but in them APP14 is clearly right, so make it win.
+            if (app14Segment != null && app14Segment.isAdobeJpegSegment()) {
+                final int colorTransform = app14Segment.getAdobeColorTransform();
+                switch (colorTransform) {
+                case App14Segment.ADOBE_COLOR_TRANSFORM_UNKNOWN:
+                    if (numberOfComponents == 3) {
+                        colorType = ImageInfo.ColorType.RGB;
+                    } else if (numberOfComponents == 4) {
+                        colorType = ImageInfo.ColorType.CMYK;
+                    }
+                    break;
+                case App14Segment.ADOBE_COLOR_TRANSFORM_YCbCr:
+                    colorType = ImageInfo.ColorType.YCbCr;
+                    break;
+                case App14Segment.ADOBE_COLOR_TRANSFORM_YCCK:
+                    colorType = ImageInfo.ColorType.YCCK;
+                    break;
+                default:
+                    break;
                 }
-                break;
-            case App14Segment.ADOBE_COLOR_TRANSFORM_YCbCr:
-                colorType = ImageInfo.ColorType.YCbCr;
-                break;
-            case App14Segment.ADOBE_COLOR_TRANSFORM_YCCK:
-                colorType = ImageInfo.ColorType.YCCK;
-                break;
-            default:
-                break;
-            }
-        } else if (jfifSegment != null) {
-            if (numberOfComponents == 1) {
-                colorType = ImageInfo.ColorType.GRAYSCALE;
-            } else if (numberOfComponents == 3) {
-                colorType = ImageInfo.ColorType.YCbCr;
-            }
-        } else {
-            switch (numberOfComponents) {
-            case 1:
-                colorType = ImageInfo.ColorType.GRAYSCALE;
-                break;
-            case 2:
-                colorType = ImageInfo.ColorType.GRAYSCALE;
-                transparent = true;
-                break;
-            case 3:
-            case 4:
+            } else if (jfifSegment != null) {
+                if (numberOfComponents == 1) {
+                    colorType = ImageInfo.ColorType.GRAYSCALE;
+                } else if (numberOfComponents == 3) {
+                    colorType = ImageInfo.ColorType.YCbCr;
+                }
+            } else {
                 boolean have1 = false;
                 boolean have2 = false;
                 boolean have3 = false;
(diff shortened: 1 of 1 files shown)


kinow:
Thanks for the pull request and sorry for taking so long to review it @jephillips34 . Will try to get on it in the next days :+1:

kinow:
Rebased onto master and squashed commits. Let's start reviewing it :)

src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
```diff
@@ -824,44 +824,44 @@ public ImageInfo getImageInfo(final ByteSource byteSource, final Map<String, Obj
...
-                colorType = ImageInfo.ColorType.GRAYSCALE;
-                break;
-            case 2:
-                colorType = ImageInfo.ColorType.GRAYSCALE;
-                transparent = true;
-                break;
-            case 3:
-            case 4:
+            } else if (jfifSegment != null) {
+                if (numberOfComponents == 1) {
```
kinow, line 859: This condition is now impossible since we are in the switch-case for `numberOfComponents == 3 | 4`.

I think the previous logic was closer to the documentation from [Oracle on Java/colors metadata](https://docs.oracle.com/javase/6/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html#color):

> If a JFIF APP0 marker segment is present, the colorspace is known to be either grayscale or YCbCr. If an APP2 marker segment containing an embedded ICC profile is also present, then the YCbCr is converted to RGB according to the formulas given in the JFIF spec, and the ICC profile is assumed to refer to the resulting RGB space. 

@jephillips34 what about this fix (diff from `master`)?

```diff
diff --git a/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java b/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
index 288c2398..1cc0fcae 100644
--- a/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
@@ -834,6 +834,8 @@ public class JpegImageParser extends ImageParser implements XmpEmbeddable {
                     colorType = ImageInfo.ColorType.RGB;
                 } else if (numberOfComponents == 4) {
                     colorType = ImageInfo.ColorType.CMYK;
+                } else if (numberOfComponents == 1) {
+                    colorType = ImageInfo.ColorType.GRAYSCALE;
                 }
                 break;
             case App14Segment.ADOBE_COLOR_TRANSFORM_YCbCr:
diff --git a/src/test/java/org/apache/commons/imaging/formats/jpeg/specific/JpegImageParserTest.java b/src/test/java/org/apache/commons/imaging/formats/jpeg/specific/JpegImageParserTest.java
index 3f86af4d..b2ec1d52 100644
--- a/src/test/java/org/apache/commons/imaging/formats/jpeg/specific/JpegImageParserTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/jpeg/specific/JpegImageParserTest.java
@@ -22,7 +22,9 @@ import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
 
+import org.apache.commons.imaging.ImageInfo;
 import org.apache.commons.imaging.ImageReadException;
+import org.apache.commons.imaging.common.bytesource.ByteSource;
 import org.apache.commons.imaging.common.bytesource.ByteSourceFile;
 import org.apache.commons.imaging.formats.jpeg.JpegImageParser;
 import org.apache.commons.imaging.formats.jpeg.decoder.JpegDecoderTest;
@@ -49,4 +51,18 @@ public class JpegImageParserTest {
         assertEquals(-16777216, image.getRGB(0, 0));
         assertEquals(-12177367, image.getRGB(198, 13));
     }
+
+    /**
+     * For IMAGING-310. Verify that a JPEG image, with an APP14 (Adobe segment) present,
+     * and 1 component must return the image color type as grayscale.
+     */
+    @Test
+    public void testGrayscaleImageWithApp14Segment() throws IOException, ImageReadException {
+        final File imageFile = new File(
+                JpegDecoderTest.class.getResource("/images/jpeg/IMAGING-310/sample-grayscale-with-app14segment.jpg")
+                        .getFile());
+        final JpegImageParser parser = new JpegImageParser();
+        ImageInfo imageInfo = parser.getImageInfo(new ByteSourceFile(imageFile));
+        assertEquals(ImageInfo.ColorType.GRAYSCALE, imageInfo.getColorType());
+    }
 }
```

The test uses the image from your JIRA issue. The test fails without the patch above. If you are happy with this, feel free to update this branch and it should be ready to be merged :+1: 

Thanks!
Bruno

kinow (changes requested):
@jephillips34 I'm still

src/main/java/org/apache/commons/imaging/formats/jpeg/JpegImageParser.java
jephillips34: Bruno, thank you for reviewing this.

I agree that your much smaller change, simply adding the check for only one component inside the handling of ADOBE_COLOR_TRANSFORM_UNKNOWN would allow the colorType of my test image to be assigned correctly as grayscale.  It would not cover the case of a 2 component Grayscale image with alpha channel, but that could be easily added as an additional condition in that same code block.  However, that would leave us with 3 separate blocks of code, all checking for one or two components, and setting those to grayscale.  

I thought it would be more correct to pull that common element out of those three code blocks, and only leave the special-case  appSegment handling for the 3- and 4-component cases, where they might actually affect the colorType determination.

Also, it just makes sense to me to first look at number of components, given the nature of the Grayscale colorType....by definition, an image with only one color component is Grayscale...it could not be anything else, regardless of any appSegment details.

I did leave the check for one component in the jfifSegment code block, though I realize that the condition is now impossible, and I would be agreeable to removing that.

What do you think?
-John

jephillips34: Bruno, I forgot to thank you for writing and setting up the test case.
Thank you!
-John

kinow: Hi @jephillips34 

>Bruno, thank you for reviewing this.
>Bruno, I forgot to thank you for writing and setting up the test case.

Thank **you** for raising the issue and providing a PR. It's extremely helpful.

>I agree that your much smaller change, simply adding the check for only one component inside the handling of ADOBE_COLOR_TRANSFORM_UNKNOWN would allow the colorType of my test image to be assigned correctly as grayscale. It would not cover the case of a 2 component Grayscale image with alpha channel, but that could be easily added as an additional condition in that same code block.

:+1: 

>However, that would leave us with 3 separate blocks of code, all checking for one or two components, and setting those to grayscale.

Yeah, I preferred your version, but I think a unit test failed, and then that led me to reading the specs to remind me of why there was that `if (jfifSegment != null)` block. Maybe we could find a way to accommodate the fix for the issue, and also improve the code.

>I thought it would be more correct to pull that common element out of those three code blocks, and only leave the special-case appSegment handling for the 3- and 4-component cases, where they might actually affect the colorType determination.
>
>Also, it just makes sense to me to first look at number of components, given the nature of the Grayscale colorType....by definition, an image with only one color component is Grayscale...it could not be anything else, regardless of any appSegment details.
>
>I did leave the check for one component in the jfifSegment code block, though I realize that the condition is now impossible, and I would be agreeable to removing that.

We need to handle this case:

>If a JFIF APP0 marker segment is present, the colorspace is known to be either grayscale or YCbCr. If an APP2 marker segment containing an embedded ICC profile is also present, then the YCbCr is converted to RGB according to the formulas given in the JFIF spec, and the ICC profile is assumed to refer to the resulting RGB space.

So if we are able to keep the existing logic, have all unit tests to pass, then we can definitely simplify the code.

At the moment I believe we have three cases for choosing the color type in the `JpegImageParser`. The first being when you have an App14 segment and it's an Adobe JPEG. The color type is chosen using the Adobe color transform from that segment.

The second handles the case when a JFIF segment is present, using just the number of components.

The third and final case has special logic for choosing color type & alpha based on number of components, SOFn segments, etc.

From a quick look this morning, I think there 1st case needs to stay the way it's, as it's related to the Adobe JPEG. The second and third cases are a bit more similar, but if you look at the switch/if statements, it's different how each case is handling three components for instance.

So I am not sure if we would be able to have 1 block handling Adobe JPEG, JFIF, and the last case.

But feel free to update this PR with code if you have any suggestions. It might be easier to discuss with some example code to see how we can improve (IMO one big improvement would be split that method into smaller methods, making it easier to distinguish these cases of Adobe, JFIF, etc.)

Thanks!
Bruno

kinow: Ping 2. Release going out soon :steam_locomotive:

garydgregory:
Closing: No reply.

PR #177, opened 2021-10-18
IMAGING-314: Add GeoTIFF metadata to output

gwlucastrig:
This PR extends the current example application that shows how to extract tags and metadata from a TIFF file using the Apache Commons Imaging API.  The new logic includes a summary of GeoTIFF related elements. Some example output is given on the JIRA issue associated with this change.


kinow:
Created placeholder issue https://issues.apache.org/jira/browse/IMAGING-314

kinow:
Merged. Thanks @gwlucastrig !

PR #183, opened 2021-11-06
Fail-fast on null inputs in AbstractPixelParser constructor (oss-fuzz 34180)

kinow:
The code is still throwing a `NullPointerException`, but now it will throw the exception in the constructor, instead of only when a method of the object is called.


garydgregory:
Closing: Redoing since the class name has changed.

kinow:
Thank you!

PR #194, opened 2022-01-07
Simplify conditions and avoid extra computations.

arturobernalg:


src/main/java/org/apache/commons/imaging/palette/PaletteFactory.java
```diff
@@ -359,7 +354,7 @@ public Palette makeQuantizedRgbPalette(final BufferedImage src, final int max) {
             LOGGER.finest("width * height: " + (width * height));
         }
 
-        subsets = divide(subsets, max, table, precision);
+        divide(subsets, max, table, precision);
```
kinow, line 357: Can you also modify `divide` to it returns `void` @arturobernalg, please? Just replace the two `return variable` statement by a simple `return` :+1: 

Thank you!

arturobernalg: HI @kinow 
Done.

kinow (approved):

kinow:
Merged, thank you!

PR #196, opened 2022-01-22
IMAGING-325: Prevent OOM error in BmpImageParser

kinow:


src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
```diff
@@ -385,6 +385,10 @@ private BmpImageContents readImageContents(final InputStream is,
                     + bhi.compression);
         }
 
+        if (paletteLength < 0) {
+            throw new ImageReadException("Invalid negative palette length: " + paletteLength);
```
garydgregory, line 389: The exception message above this one uses a "BMP" prefix, should this message do so as well?

garydgregory: Also, isn't the underlying issue that the colorTableSize is negative? Shouldn't we check that instead?

garydgregory: Hm, and that one is negative because bhi.bitsPerPixel is greater than should be allowed, right?

garydgregory: So the BmpHeaderInfo should not have been created in the first place? Maybe that's where the error checking should really go.

kinow: Good catch on the exception message prefix, thanks!

>Also, isn't the underlying issue that the colorTableSize is negative? Shouldn't we check that instead?

I thought there would be a negative number before, so used the provided BMP to test and noticed it was due to integer overflow when multiplying values.

>Hm, and that one is negative because bhi.bitsPerPixel is greater than should be allowed, right?

`bhi.bitsPerPixel` is `8`, which looks OK.

The `BmpHeaderInfo` contains the number of colors used, and in this case is `1667760130`, below the limit for Java integer. But when multiplied by `4` (due to compression type) we get `6671040520`, causing the integer overflow.

image

>So the BmpHeaderInfo should not have been created in the first place? Maybe that's where the error checking should really go.

Good question, I also wondered where would be the best place to put this check. I couldn't figure out a way to prevent it from happening unless the numbers were multiplied and then we could check if it's negative. That's why I left the check here, below the `switch` statement, so that it covers all possible branches checking for the integer overflow.

garydgregory: Thanks for the clarification. I still think the problem can be detected a touch earlier and we can give a better exception message. If the * 4 operation would result in a negative number, then the problem is the operand of that * 4.

kinow: Maybe we can write some helper functions to identify when an overflow will occur? Similar to these ones: https://wiki.sei.cmu.edu/confluence/display/java/NUM00-J.+Detect+or+prevent+integer+overflow

Adding a TODO to replace by [multiplyExact](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Math.html#multiplyExact(int,int)) when we move to Java 11?

If that looks good, I'd prefer to fix this issue as-is, and create a separate issue for this (with a mention to this PR), since there are lots of other places that would use these functions. WDYT?

garydgregory: That sounds great. I am ok with updating this component to Java 11.

kinow: Yeah me too. Let's do it before the 1.0 release. Thanks Gary!

kinow: Follow up issue: https://issues.apache.org/jira/browse/IMAGING-326

Created as a blocker for 1.0, but will try to include in 1.0-alpha3 if possible, to let users experiment with it. Thanks Gary, merging & closing this issue.

PR #202, opened 2022-02-06
Extract same conditions and reuse it.

arturobernalg:


kinow (commented):
I hold back on changes like this in Imaging since a lot of these changes are micro-improvements that may have the same bytecode generated anyway, and the majority of the code in Imaging is written a little awkwardly because it's ported from specifications, books, C code, etc. And I find it better to have the code closer to other implementations for troubleshooting/comparing the code, unless the change has a significant improvement for readability (e.g. Java 11, Java 16 features, etc) or performance.

PR #209, opened 2022-03-09
IMAGING-306: Constants in ColorHsv are not in tune with ColorConversions algorithms

oskaredv:
We have worked on [IMAGING](https://issues.apache.org/jira/browse/IMAGING-306) as part of a school assignment


garydgregory:
Hi @oskaredv 
Run the build locally before pushing a PR using the default goal: `mvn` as this PR breaks the build:
```
Error:  src/main/java/org/apache/commons/imaging/color/ColorConversions.java:[182] (regexp) RegexpSingleline: Line has trailing spaces.
Error:  src/main/java/org/apache/commons/imaging/color/ColorConversions.java:[207] (regexp) RegexpSingleline: Line has trailing spaces.
```

PR #210, opened 2022-03-29
Imaging 330: Add PNG predictor to reduce output size

gwlucastrig:
This change enhances the PngWriter class and PngImagingParameters class to allow the use of predictors. This change should reduce the size of output images written in PNG format. The resulting images will carry exactly the same data. There will be no loss of pixels or image quality. But the results will be smaller than those currently produced by either Imaging or Java's ImageIO class.

The default setting for PngImagingParameters is currently to not apply a predictor. So it is consistent with the existing behavior.

I am going to let this run through the build process.  When I see how the coverage analysis plays out, I will write a new JUnit test (if necessary) and submit it as part of this Pull Request.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/png/PngImagingParameters.java
+++ b/src/main/java/org/apache/commons/imaging/formats/png/PngImagingParameters.java
@@ -37,6 +37,8 @@ public class PngImagingParameters extends XmpImagingParameters {
 
     private boolean forceTrueColor = false;
 
+    private boolean predictorEnabled = false;
+
     /**
      * Used in write operations to indicate the Physical Scale - sCAL.
      *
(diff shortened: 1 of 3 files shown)


gwlucastrig:
This completes the work.  Unfortunately, there may have been a problem when I squashed the commits.  If there is a problem with the pull request, let me know and I will simply cancel this pull request and start over.

kinow (approved):
Looks good to me! @gwlucastrig I'll squash the commits and update JIRA and changes.xml. Do you want to check coverage? Looks like the bot didn't comment in this PR? Thanks!

kinow:
Commits squased, changes.xml entry added, branch merged, and JIRA updated. Thanks @gwlucastrig !

PR #212, opened 2022-04-07
Design Pattern Implement

paras160500:
Design pattern implemented

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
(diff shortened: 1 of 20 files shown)


garydgregory:
Remove all of the idea files.

kinow (changes requested):
The idea files need to be removed as @garydgregory pointed out (you can add to your `~/.gitignore` to globally ignore it, if you'd like).

Also, would be nice to have an explanation on the motivation for the change. Classes like `ScanExpediter` are used internally only, unless the user has a **very** specific case to modify or use it. So I can't really see how the change would benefit users or maintainers.

garydgregory:
Closing: No reply from the author.

PR #214, opened 2022-04-09
Simplify conditional expression and reuse variables.

arturobernalg:


src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
```diff
@@ -374,16 +374,14 @@ private void readMCU(final JpegInputStream is, final int[] preds, final Block[]
                                 is,
                                 huffmanACTables[scanComponent.acCodingTableSelector]);
                         final int ssss = rs & 0xf;
-                        final int rrrr = rs >> 4;
-                        final int r = rrrr;
```
kinow, line 378: If others prefer to have this expressions replaced, that'd be OK for me. But I prefer to keep they as they are, since if there's any performance gain, it might be nanoseconds.

Similar to [this comment](https://github.com/apache/commons-imaging/pull/155#discussion_r662990219) from another PR @arturobernalg, these variables are present in code in other languages, and they all come from documents/specification/revisions/etc.

Looking at this one, it looks like the original author took care to reproduce a flow chart from [this document](https://www.w3.org/Graphics/JPEG/itu-t81.pdf).

image

I prefer to have the code as similar to the original spec/paper/blog/code/etc, since it's normally easier to work on it if I have to re read the docs, or look at code in other languages.

Imaging has a lot of that, in byte shifting operations, and other places where unless it's a really bad code pattern, or performs drastically worst, I prefer to leave the code to be similar to the source of origin.

Bruno

kinow (changes requested):

src/main/java/org/apache/commons/imaging/formats/jpeg/decoder/JpegDecoder.java
arturobernalg: HI @kinow 
true. I completely forgot. 
Reverted.
TY

PR #268, opened 2023-01-12
[IMAGING-342] Read PNG metadata from iTXt chunk

Glavo:
In [PNG specification](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.Anc-text), `iTXt` chunk is semantically equivalent to the `tEXt` and `zTXt` chunks, but `PngImageParser::getMetadata` does not recognize the `iTXt` chunk.

This PR fixes this problem.


kinow (commented):
:+1: small change, but do you think we could have a unit test for this (to make sure we don't accidentally drop this from the parser again).

We also need a JIRA to add it to the changelog. Can you create one, please, @Glavo ?

Thanks!

kinow:
I also realized it's hard to know what version of each spec we support, so here's a new one for 1.0 - https://issues.apache.org/jira/browse/IMAGING-341

Glavo:
I added a test to check whether the metadata with Unicode characters can be correctly obtained from PNG images.

kinow (approved):
Confirmed the test was using that chunk type.

image

CI happy, looks good to me! I will squash the commits into a single one prefixed by the [IMAGING-...] issue ID, and then merge it later.

Thanks @Glavo !

kinow:
Merged!

PR #270, opened 2023-01-16
Fix unit test

kpouer:
Fix two unit tests that were broken.
PixelDensityRoundtrip was now writing data in a byte array instead of a tmp file, but was trying to read the tmp file anyway.

And

PngWriteForceTrueColorText that was writing a png into a file named .gif, but was unable to read it because the extension was checked and refused by png parser.


kinow (approved):
Thank you!

PR #271, opened 2023-01-19
[IMAGING-345] Unit test : make tests in memory instead of writing hundred of tmp files

kpouer:
An attempt to perform unit tests in memory using byte array when possible instead of writing temporary files.
Tests pass much faster and do not make useless write on my SSD.
I know that having the tmp files might be useful in some case, if you  think it it necessary I could update the patch to write tmp file in case of failed test.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/formats/bmp/BmpRoundtripTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/bmp/BmpRoundtripTest.java
@@ -22,13 +22,10 @@ import org.apache.commons.imaging.ImageReadException;
 import org.apache.commons.imaging.ImageWriteException;
 import org.apache.commons.imaging.Imaging;
 import org.apache.commons.imaging.internal.Debug;
-import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.Test;
 
 import java.awt.image.BufferedImage;
-import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.util.Random;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -139,9 +136,6 @@ public class BmpRoundtripTest extends BmpBaseTest {
         final BufferedImage srcImage = imageDataToBufferedImage(rawData);
 
         final byte[] bytes = Imaging.writeImageToBytes(srcImage, ImageFormats.BMP);
-
-        final File tempFile = Files.createTempFile("temp", ".bmp").toFile();
-        FileUtils.writeByteArrayToFile(tempFile, bytes);
 
         final BufferedImage dstImage = Imaging.getBufferedImage(bytes);
 
--- a/src/test/java/org/apache/commons/imaging/formats/png/PngTextTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/png/PngTextTest.java
@@ -24,8 +24,6 @@ import org.junit.jupiter.api.Test;
 import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -73,9 +71,6 @@ public class PngTextTest extends PngBaseTest {
             pngImageParser.writeImage(srcImage, baos, writeParams);
             bytes = baos.toByteArray();
         }
-
-        final File tempFile = Files.createTempFile("temp", ".png").toFile();
-        FileUtils.writeByteArrayToFile(tempFile, bytes);
 
         final PngImageInfo imageInfo = (PngImageInfo) Imaging.getImageInfo(bytes);
         assertNotNull(imageInfo);
--- a/src/test/java/org/apache/commons/imaging/formats/png/PngWriteForceTrueColorText.java
+++ b/src/test/java/org/apache/commons/imaging/formats/png/PngWriteForceTrueColorText.java
@@ -21,9 +21,8 @@ import org.apache.commons.imaging.internal.Debug;
 import org.junit.jupiter.api.Test;
 
 import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileOutputStream;
-import java.nio.file.Files;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -50,13 +49,14 @@ public class PngWriteForceTrueColorText extends PngBaseTest {
                 final BufferedImage image = pngImageParser.getBufferedImage(imageFile, new PngImagingParameters());
                 assertNotNull(image);
 
-                final File outFile = Files.createTempFile(imageFile.getName() + ".", ".png").toFile();
                 // Debug.debug("outFile", outFile);
 
                 final PngImagingParameters params = new PngImagingParameters();
-                params.setForceTrueColor(Boolean.TRUE);
-                try (FileOutputStream fos = new FileOutputStream(outFile)) {
-                    pngImageParser.writeImage(image, fos, params);
+                byte[] outFile;
+                try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+                    params.setForceTrueColor(Boolean.TRUE);
+                    pngImageParser.writeImage(image, os, params);
+                    outFile = os.toByteArray();
                 }
 
                 final BufferedImage image2 = pngImageParser.getBufferedImage(outFile, new PngImagingParameters());
--- a/src/test/java/org/apache/commons/imaging/roundtrip/FormatInfo.java
+++ b/src/test/java/org/apache/commons/imaging/roundtrip/FormatInfo.java
@@ -130,4 +130,16 @@ class FormatInfo {
         this.identicalSecondWrite = identicalSecondWrite;
         this.preservesResolution = preservesResolution;
     }
+
+    @Override
+    public String toString() {
+        return "FormatInfo{" +
+            "format=" + format +
+            ", canRead=" + canRead +
+            ", canWrite=" + canWrite +
+            ", colorSupport=" + colorSupport +
+            ", identicalSecondWrite=" + identicalSecondWrite +
+            ", preservesResolution=" + preservesResolution +
+            '}';
+    }
 }
--- a/src/test/java/org/apache/commons/imaging/roundtrip/NullParametersRoundtripTest.java
+++ b/src/test/java/org/apache/commons/imaging/roundtrip/NullParametersRoundtripTest.java
@@ -22,8 +22,8 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import java.awt.image.BufferedImage;
-import java.io.File;
-import java.nio.file.Files;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -38,13 +38,17 @@ public class NullParametersRoundtripTest extends RoundtripBase {
     @MethodSource("data")
     public void testNullParametersRoundtrip(final FormatInfo formatInfo) throws Exception {
         final BufferedImage testImage = TestImages.createFullColorImage(1, 1);
-        final File temp1 = Files.createTempFile("nullParameters.", "." + formatInfo.format.getDefaultExtension()).toFile();
-        Imaging.writeImage(testImage, temp1, formatInfo.format);
-        Imaging.getImageInfo(temp1);
-        Imaging.getImageSize(temp1);
-        Imaging.getMetadata(temp1);
-        Imaging.getICCProfile(temp1);
-        final BufferedImage imageRead = Imaging.getBufferedImage(temp1);
+        final byte[] temp1;
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+            Imaging.writeImage(testImage, byteArrayOutputStream, formatInfo.format);
+            temp1 = byteArrayOutputStream.toByteArray();
+        }
+        final String filename = "nullParameters." + formatInfo.format.getDefaultExtension();
+        Imaging.getImageInfo(new ByteArrayInputStream(temp1), filename);
+        Imaging.getImageSize(new ByteArrayInputStream(temp1), filename);
+        Imaging.getMetadata(new ByteArrayInputStream(temp1), filename);
+        Imaging.getICCProfile(new ByteArrayInputStream(temp1), filename);
+        final BufferedImage imageRead = Imaging.getBufferedImage(new ByteArrayInputStream(temp1), filename);
 
         assertNotNull(imageRead);
     }
--- a/src/test/java/org/apache/commons/imaging/roundtrip/RoundtripBase.java
+++ b/src/test/java/org/apache/commons/imaging/roundtrip/RoundtripBase.java
@@ -21,17 +21,15 @@ import org.apache.commons.imaging.ImageReadException;
 import org.apache.commons.imaging.ImageWriteException;
 import org.apache.commons.imaging.ImagingParameters;
 import org.apache.commons.imaging.common.RgbBufferedImageFactory;
-import org.apache.commons.imaging.internal.Debug;
 import org.apache.commons.imaging.internal.ImageParserFactory;
 import org.junit.jupiter.params.provider.Arguments;
 
 import java.awt.image.BufferedImage;
-import java.io.File;
-import java.io.FileOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.util.stream.Stream;
 
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class RoundtripBase {
@@ -39,15 +37,14 @@ public class RoundtripBase {
     protected void roundtrip(final FormatInfo formatInfo, final BufferedImage testImage,
                              final String tempPrefix, final boolean imageExact) throws IOException,
             ImageReadException, ImageWriteException {
-        final File temp1 = Files.createTempFile(tempPrefix + ".", "."
-                + formatInfo.format.getDefaultExtension()).toFile();
-        Debug.debug("tempFile: " + temp1.getName());
 
         final ImageParser imageParser = ImageParserFactory.getImageParser(formatInfo.format);
 
         final ImagingParameters params = ImageParserFactory.getImageParser(formatInfo.format).getDefaultParameters();
-        try (FileOutputStream fos = new FileOutputStream(temp1)) {
-            imageParser.writeImage(testImage, fos, params);
+        byte[] temp1;
+        try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
+            imageParser.writeImage(testImage, bos, params);
+            temp1 = bos.toByteArray();
         }
 
         final ImagingParameters readParams = ImageParserFactory.getImageParser(formatInfo.format).getDefaultParameters();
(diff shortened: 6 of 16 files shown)


kinow (commented):
I'm busy with $work these days, but

>An attempt to perform unit tests in memory using byte array when possible instead of writing temporary files.
>Tests pass much faster and do not make useless write on my SSD.
>I know that having the tmp files might be useful in some case, if you think it it necessary I could update the patch to write tmp file in case of failed test.

I think it could be useful in some cases, like if you want to debug/troubleshoot it, but I think speeding up the tests and reducing IO sound like a good idea. In case a test fails and we need to troubleshoot, it shouldn't be too hard to write it to disk to investigate (assuming that's needed at all). So I'm a big +1 for this, @kpouer .

I had a quick look and looks like you are adding some public methods too, in non-test files. Were these necessary for the change in this PR, @kpouer ?

Final note, it'd be necessary to squash the commits later. If you are still sorting out how to organize the change, feel free to keep it as is until you have it ready. Otherwise, squash it so that it's ready to be merged, please.

Thanks heaps!

garydgregory:
Keep in mind that tests must be able to run in containers that may not provide much memory.

kpouer:
Hi @kinow,
About the new methods, it is utility that get various information about the image. Some format are not detected only from their data and the file extension is necessary. It was of course working with tmp files, but not with byte arrays.
I took inspiration of the equivalent method taking an InputStream and a filename, for example
There was a 

public static ICC_Profile getICCProfile(final InputStream is, final String fileName)

Then I created
public static ICC_Profile getICCProfile(final String fileName, final byte[] bytes) (the parameter order is not the same but it follows the constructor parameter orders of the ByteSourceArray class (however when rereading I think that Imaging class should always have the same parameter orders, it would be easier to understand.
I could have used getICCProfile(ByteSource) but the method was protected then I should have changed it's visibility.

About the troubleshooting I'll update the branch in order to save tmp files for failed test.

@garydgregory I don't think it will take a lot of memory as images are processed one by one, I think it should store something like 2 or 3 time the same image during some tests, then the memory is released to the GC to process the next image. And test images do not seem to be huge. However I am interested but not sure about what you mean by tested in containers. Do I have a way to test it to validate ?

garydgregory:
I was referring to GitHub Actions builds being resource constrained (to what extent I don't know exactly).

kpouer:
Hey, I was able to do the same test without changing public api so I reverted the change (even if I think those functions might be useful, it could be done in another PR if necessary.
I added code to save tmp file in case of error during test (catched throwable, I don't like it but had no idea how to do it in another way). For memory I tried to isolate ByteArrayOutputStream in blocks to allow GC to collect them as soon as possible, I hope it helps.
when you think it is ok I can resubmit the same patch in only one commit (or update it if you think there are more changes to do

kinow:
Thanks @kpouer can you squash your commits? And would you be able to create a JIRA for this? Even though it's a change in the tests I'd like to review (will try to compare number of IO ops, and memory on my laptop at least) and include it in the changelog for the next release. Thanks!

kinow (commented):
Thanks @kpouer can you squash your commits? And would you be able to create a JIRA for this? Even though it's a change in the tests I'd like to review (will try to compare number of IO ops, and memory on my laptop at least) and include it in the changelog for the next release. Thanks!

kpouer:
Hey, I was not sure how I could squash commits without deleting my branch and pushing it again so I created a new branch and a new pull request (same code but squashed into one commit)
https://github.com/apache/commons-imaging/pull/272
I would be happy to create a JIRA ticket but I don't have an account on it. It says I could send an email to private@commons-imaging.apache.org if that email is setup. Is it ?

kinow:
Hi @kpouer 

> Hey, I was not sure how I could squash commits without deleting my branch and pushing it again so I created a new branch and a new pull request (same code but squashed into one commit) #272

It's probably useful to get used to squashing commits, both for Apache Commons code contributions, but for other OSS that follow similar approach. The issue with opening new PR's is that a) you may have to add more commits again, and then squash again, and b) it gets more complicated to keep track of discussions and reviews (there's also c) where users/devs subscribed to the repo may get spammed by similar changes).

If you'd like you can try on this PR, as it's still open: https://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

I squash rebasing interactively. Basically what I'd do here would be something similar to:

```bash
git fetch --all
git rebase upstream/master # upstream for me is apache/commons-imaging, kinow for me is the kinow/commons-imaging
gitk # or git log -n N, but I like to take a look at the GUI too, then count the number of commits since the top of master
# you have 17 commits here, ATOW
git rebase -i HEAD~17  # -i is interactive, i.e. I will interact with git to decide how to rebase
# each commit is prefixed by an action code, I mark the first line as r-eword
# and for each new commit line in the VIM/editor, I mark as f-ixup
```

image

When I save my editor, `git rebase` proceeds, opening the editor again as I asked to reword the fix commit. Now I write a good commit line. If I already have the JIRA number I prefix the commit message with [IMAGING-NNNN] Perform tests in memory instead of disk, making them faster".

When I save this file again, git will use the text as the commit message of the reworded commit. The other commits have been all "fixed-up". So now everything has been "squashed" into a single commit.

At this point I `gitk` again, or `git log -n 1` or `-n 2` to look at the commit history again.

image

At this point your local branch contains the squashed commits for your change. You can browse the code, run `mvn`, etc., until you are sure that everything is in order. Then you are ready to `push -f kinow/branch-name` (where kinow will be your personal remote, and branch-name must be `unit_test_tmp_fil....`, I think).

And the pull request will have been updated with your forced (-f) commit, showing a single commit for reviewers :+1: 

If that works you can close your other PR. Or if you prefer to try this later, feel free to close this issue :+1: up to you.

> I would be happy to create a JIRA ticket but I don't have an account on it. It says I could send an email to [private@commons-imaging.apache.org](mailto:private@commons-imaging.apache.org) if that email is setup. Is it ?

I am sorry, due to bots SPAMing our JIRA server we had to lock it for new users :disappointed_relieved:

Send the email, please, to the commons private mail (it's the email address you mentioned above, minus the "-imaging" part) with the info on https://infra.apache.org/jira-guidelines.html#account, and I or another committer will create your account.

Really sorry for the inconvenience :bow: 

Thanks @kpouer !

kinow:
>  kpouer force-pushed the unit_test_tmp_file_optimization branch from eddd5bf to 1947c86 

:clap: great job! Much easier to review, and thanks for closing the other PR. Bravo @kpouer :+1:

kpouer:
Yes I was able to squash but was missing the "force" option for the push

kinow:
Sorry for the delay, next on my list! :nerd_face:

src/test/java/org/apache/commons/imaging/roundtrip/RoundtripBase.java
```diff
@@ -39,35 +38,52 @@ public class RoundtripBase {
...
-                + formatInfo.format.getDefaultExtension()).toFile();
-        Debug.debug("tempFile: " + temp1.getName());
 
         final ImageParser imageParser = Util.getImageParser(formatInfo.format);
 
+        byte[] temp1;
         final ImagingParameters params = Util.getImageParser(formatInfo.format).getDefaultParameters();
-        try (FileOutputStream fos = new FileOutputStream(temp1)) {
-            imageParser.writeImage(testImage, fos, params);
+        try (ByteArrayOutputStream bos = new ByteArrayOutputStream(1000000)) {
```
kinow, line 46: Maybe leave it the default value and let it handle expanding the internal buffer? That could be helpful if we load many smaller images, I think (although I have no idea the distribution of sizes of our images & their metadata too).

src/test/java/org/apache/commons/imaging/formats/bmp/BmpRoundtripTest.java
```diff
@@ -140,17 +139,21 @@ private void writeAndReadImageData(final int[][] rawData) throws IOException, Im
...
+            assertNotNull(dstImage);
+            assertEquals(srcImage.getWidth(), dstImage.getWidth());
+            assertEquals(srcImage.getHeight(), dstImage.getHeight());
+
+            final int[][] dstData = bufferedImageToImageData(dstImage);
+            compare(rawData, dstData);
+        } catch (final Throwable e) {
+            final Path tempFile = Files.createTempFile("temp", ".bmp");
+            Files.write(tempFile, bytes);
+            System.err.println("Failed tempFile " + tempFile);
```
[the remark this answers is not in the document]
kinow: Also, reading again the code, I think we can rethink this fallback mechanism.

Right now, in case there's any exception of any type, it writes the image to the disk. But that might end up doing that unnecessarily for some exception types.

It might be simple to just throw the error and let a developer call the code to write it to the disk, I think. Otherwise we will also have to remember to use this boilerplace code whenever we write new test code that loads images into the memory.

Sorry @kpouer, as I believe your initial PR didn't have this, and you added it based on our initial feedback. Right now I'm included to undo that, but give me some time to think about this one :+1:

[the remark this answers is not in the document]
kinow: (and no need to worry, I can undo the change if we decide to go down that path :wink: :+1: )

kinow:
Used Docker to run the build from this PR limiting the allowed memory for the container. The build crashed at 64mb, 128mb, 256mb. And it passed for 512mb and higher.

Switched to `master`, and 512mb passes, and 256mb fails as well. Command line used:

```bash
docker run -it --rm --name my-maven-project -v "$(pwd)":/usr/src/mymaven -w /usr/src/mymaven --memory=256m maven:3.8-openjdk-17 mvn clean install
```

src/test/java/org/apache/commons/imaging/formats/bmp/BmpRoundtripTest.java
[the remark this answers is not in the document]
kpouer: Hi, in fact I could undo the change that automatically write the tmp file as it is probably useless in case of automatic tests. But add back the code that writes the file but commented out so it could be activated manually by a developper if needed ?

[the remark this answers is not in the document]
kinow: >Hi, in fact I could undo the change that automatically write the tmp file as it is probably useless in case of automatic tests. But add back the code that writes the file but commented out so it could be activated manually by a developper if needed ?

That'd be useful! I really think that'd be better than just writing whenever there's an exception. Sorry for the extra work @kpouer !

kpouer:
I updated the branch, first I wanted to comment out code to create tmp files for troubleshooting but finally I just did enclose it in a if (false) {} block, this way no problem with imports and it is even easiest to set it to true for troubleshooting.
I'm exploring buiding with docker just to see if something can be done more efficiently

src/test/java/org/apache/commons/imaging/formats/png/PngTextTest.java
```diff
@@ -74,8 +76,10 @@ public void test() throws Exception {
             bytes = baos.toByteArray();
         }
 
-        final File tempFile = Files.createTempFile("temp", ".png").toFile();
-        FileUtils.writeByteArrayToFile(tempFile, bytes);
+        if (false) { // activate saving tmp file for troubleshooting
+            final File tempFile = Files.createTempFile("temp", ".png").toFile();
+            FileUtils.writeByteArrayToFile(tempFile, bytes);
```
kinow, line 81: `LOGGER` no used here.

src/test/java/org/apache/commons/imaging/formats/png/PngWriteForceTrueColorText.java
```diff
@@ -21,21 +21,25 @@
...
 public class PngWriteForceTrueColorText extends PngBaseTest {
+    private static final Logger LOGGER = Logger.getLogger(PngWriteForceTrueColorText.class.getName());
 
     @Test
     public void test() throws Exception {
         final PngImageParser pngImageParser = new PngImageParser();
         final List<File> images = getPngImages();
         for (final File imageFile : images) {
 
+            byte[] outFile = null;
```
kinow, line 42: This could be inside the `try` block. I thought it was declared outside for some reason, but apparently it's only used within that block?

src/test/java/org/apache/commons/imaging/formats/png/PngWriteForceTrueColorText.java
```diff
@@ -50,13 +54,19 @@ public void test() throws Exception {
...
+
+                try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+                    params.setForceTrueColor(Boolean.TRUE);
+                    pngImageParser.writeImage(image, os, params);
+                    outFile = os.toByteArray();
+                }
+
+                if (false) { // activate saving tmp file for troubleshooting
+                    final Path tempFile = Files.createTempFile(imageFile.getName() + ".", ".jpg");
+                    Files.write(tempFile, outFile);
```
kinow, line 69: `LOGGER` not used here.

src/test/java/org/apache/commons/imaging/roundtrip/RoundtripBase.java
```diff
@@ -62,10 +78,18 @@ protected void roundtrip(final FormatInfo formatInfo, final BufferedImage testIm
...
 
         if (formatInfo.identicalSecondWrite) {
-            final File temp2 = Files.createTempFile(tempPrefix + ".", "." + formatInfo.format.getDefaultExtension()).toFile();
-            try (FileOutputStream fos = new FileOutputStream(temp2)) {
-                imageParser.writeImage(image2, fos, params);
+            if (false) { // activate saving tmp file for troubleshooting
+                final File temp2 = Files.createTempFile(tempPrefix + ".", "." + formatInfo.format.getDefaultExtension()).toFile();
+                try (FileOutputStream fos = new FileOutputStream(temp2)) {
+                    imageParser.writeImage(image2, fos, params);
+                }
```
kinow, line 85: `LOGGER` not used here.

kinow (commented):
@kpouer I'd be OK removing the `if (false)` block, or leaving it commented out in one test class, or adding a `README.md` in the test directory, etc.

As more tests are added, I guess other devs may ignore copying that `if (false)` block, and it's not very hard to change the code to write to the disk in case a test is failing.

Furthermore, it's hard to be consistent with this code as it's not used (see comments where in some blocks there's a call to `LOGGER.severe`, while others are missing it.

What do you think?

kpouer:
Hi @kinow I think you are right, those blocks adds complexity to the test and I am not sure they will be so useful. In the rare case someone really needs to see the file it is finally easy to add a few lines of code to write the file. I will remove them.

kinow:
> Hi @kinow I think you are right, those blocks adds complexity to the test and I am not sure they will be so useful. In the rare case someone really needs to see the file it is finally easy to add a few lines of code to write the file. I will remove them.

Excellent @kpouer . Feel free to ignore/resolve my last review comments then. I believe most/all are about the missing `LOGGER` in the `if(false)` block, and so they shouldn't be valid any longer. Thank you!

src/test/java/org/apache/commons/imaging/roundtrip/NullParametersRoundtripTest.java
```diff
@@ -38,13 +38,17 @@ public static Stream<FormatInfo> data() {
...
+        final byte[] temp1;
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+            Imaging.writeImage(testImage, byteArrayOutputStream, formatInfo.format);
+            temp1 = byteArrayOutputStream.toByteArray();
+        }
+        final String filename = "nullParameters." + formatInfo.format.getDefaultExtension();
+        Imaging.getImageInfo(new ByteArrayInputStream(temp1), filename);
+        Imaging.getImageSize(new ByteArrayInputStream(temp1), filename);
+        Imaging.getMetadata(new ByteArrayInputStream(temp1), filename);
+        Imaging.getICCProfile(new ByteArrayInputStream(temp1), filename);
```
kinow, line 50: Hmmm, we cannot resuse the same `ByteArrayInputStream` right? :thinking: would be nicer if we closed these streams at least. Maybe we have to split it into separated test methods?

src/test/java/org/apache/commons/imaging/roundtrip/ImageAsserts.java
```diff
@@ -99,4 +99,18 @@ static void assertEquals(final File a, final File b) throws IOException {
...
+            final int bByte = 0xff & bData[i];
+
+            if (aByte != bByte) {
+                Debug.debug("i: " + i);
+                Debug.debug("aByte: " + aByte + " (0x" + Integer.toHexString(aByte) + ')');
+                Debug.debug("bByte: " + bByte + " (0x" + Integer.toHexString(bByte) + ')');
+            }
+            Assertions.assertEquals(aByte, bByte);
+        }
+    }
```
kinow, line 115: Is there a reason for using this over using JUnit's [`assertArrayEquals`](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assertions.html#assertArrayEquals-byte:A-byte:A-) ? The `Debug` is not active by default, I believe, so using this doesn't seem to add must value.

src/test/java/org/apache/commons/imaging/roundtrip/FormatInfo.java
```diff
@@ -130,4 +130,16 @@ class FormatInfo {
...
+    @Override
+    public String toString() {
+        return "FormatInfo{" +
+            "format=" + format +
+            ", canRead=" + canRead +
+            ", canWrite=" + canWrite +
+            ", colorSupport=" + colorSupport +
+            ", identicalSecondWrite=" + identicalSecondWrite +
+            ", preservesResolution=" + preservesResolution +
+            '}';
```
kinow, line 143: Looks harmless, but was there a reason for adding this one?

kpouer: It helped me a little when looking at some FormatInfo in the debugger of my IDE but if you don't find it useful I can remove that.

src/test/java/org/apache/commons/imaging/roundtrip/NullParametersRoundtripTest.java
kpouer: In fact we can reuse the same ByteArrayInputStream but the code is more verbose, and closing it has no effect (however it is useful for static analysis of IDE so they do not complains about it ;) )
try (ByteArrayInputStream bos = new ByteArrayInputStream(temp1)) {
            Imaging.getImageInfo(bos, filename);
            bos.reset();
            Imaging.getImageSize(bos, filename);
            bos.reset();
            Imaging.getMetadata(bos, filename);
            bos.reset();
            Imaging.getICCProfile(bos, filename);
            bos.reset();
            final BufferedImage imageRead = Imaging.getBufferedImage(bos, filename);
            bos.reset();
            assertNotNull(imageRead);
        }

Another option is to use the methods taking directly a byte[] but it would require to create new methods also taking the filename as the file extension is sometimes used to help reading data (unfortunately some exists but the parameter order is not consistent, sometimes the filename comes first some other time it comes after)

src/test/java/org/apache/commons/imaging/roundtrip/FormatInfo.java
kinow: Got it. Sounds like something that could be helpful to others too, and tests only, so really no harm in adding it. Thanks for the explanation!

src/test/java/org/apache/commons/imaging/roundtrip/ImageAsserts.java
kinow: I think the JUnit version should work too. Not sure if what `ImageAsserts#assertEquals(File, File)` is doing is necessary. It's reading the bytes from the stream, I assume, without handling `int` values (and thus having to apply a filter mask `0xff` to clear the other bits in the 32 byte-Java).

The point here is that if the JUnit's `assertArrayEquals` work, it's better it's less code for us to test :+1:

kinow (changes requested):
See my comment about the JUnit method, but I think this is my last review feedback. It should be ready for a final pass of :eyes: and merge. Thanks @kpouer !!!

kpouer:
Hey I missed the notification, yes a simple assertArrayEquals would be more efficient, do you want that I remove it ?

kinow:
>Hey I missed the notification, yes a simple assertArrayEquals would be more efficient, do you want that I remove it ?

Yes, if possible. Rebase and fix the conflicts, and then we can get it ready to be merged. Thanks @kpouer !

kpouer:
done

kinow (approved):
Excellent, @kpouer ! I will add a changelog and merge it. Thank you for your contribution, and for the patience!

kinow:
Merged!

PR #273, opened 2023-01-21
Unit test : Fix FullColorRoundtrip test for PCX & DCX

kpouer:
PCX & DCX have 256 color palette, so they cannot perform full color round trip on image of 300x300.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/roundtrip/FullColorRoundtrip.java
+++ b/src/test/java/org/apache/commons/imaging/roundtrip/FullColorRoundtrip.java
@@ -18,8 +18,10 @@
 package org.apache.commons.imaging.roundtrip;
 
 import java.awt.image.BufferedImage;
+import java.awt.image.RenderedImage;
 import java.util.stream.Stream;
 
+import org.apache.commons.imaging.ImageFormats;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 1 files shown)


kinow (commented):
@kpouer the `PcxImageParser` has a comment pointing to this spec: http://qzx.com/pc-gpe/pcx.txt

>24-Bit .PCX Files
>24 bit images are stored as version 5 or above as 8 bit, 3 plane images.

I opened a JPEG on GIMP, and saved as pcx. It resulted in a 1440x900 PCX image. Here's `exiftool` output:

```bash
$ exiftool orion-nebula-hd-wallpaper-1440x900.pcx 
ExifTool Version Number         : 12.40
File Name                       : orion-nebula-hd-wallpaper-1440x900.pcx
Directory                       : .
File Size                       : 2.7 MiB
File Modification Date/Time     : 2023:01:22 09:44:28+01:00
File Access Date/Time           : 2023:01:22 09:44:37+01:00
File Inode Change Date/Time     : 2023:01:22 09:44:28+01:00
File Permissions                : -rw-rw-r--
File Type                       : PCX
File Type Extension             : pcx
MIME Type                       : image/pcx
Manufacturer                    : ZSoft
Software                        : PC Paintbrush 3.0+
Encoding                        : RLE
Bits Per Pixel                  : 8
Left Margin                     : 0
Top Margin                      : 0
Image Width                     : 1440
Image Height                    : 900
X Resolution                    : 44
Y Resolution                    : 44
Color Planes                    : 3
Bytes Per Line                  : 1440
Color Mode                      : Color Palette
Image Size                      : 1440x900
Megapixels                      : 1.3
```

It's showing the 8 bits and 3 color planes. No color palette.

Passing that image through the `roundtrip` method, however, results in a `NullPointerException`. That's because, from what I could tell, the specification for these larges images state:

>24 bit images do not contain a palette.

But where the NPE happens the code is trying to access the color palette object: https://github.com/apache/commons-imaging/blob/bf31e31d98a0242c34e9bc7a5a40961f102c91cc/src/main/java/org/apache/commons/imaging/formats/pcx/PcxWriter.java#L132

But that's a separate issue (I will create one in JIRA for this).

By simply adding a `palete != null` in that line, the `roundtrip` test for that 1440x900 image completes successfully, writing the temp images and comparing them with the original ones.

So I think this PR is not needed, WDYT?

Cheers,
Bruno

kinow:
>But that's a separate issue (I will create one in JIRA for this).

https://issues.apache.org/jira/browse/IMAGING-344

kpouer:
I think you are right @kinow in fact I supposed the PCX code was good and was not aware that 24 bit PCX existed. So I close the PR, sorry for the disturbance

kinow:
>So I close the PR, sorry for the disturbance

Thank you, and not a problem at all. In the end we found a bug in the `PcxWriter` thanks to you for looking at that code :+1: keep the PR's and issues going. No matter how big or small. And if you have any questions just ping us in the dev mailing list. Cheers

PR #278, opened 2023-02-17
(Ref) #8 refactored decode image in order to reduce Cyclomatic

Darpos:
Complexity from 17 to 7 (58.8235294118 %). This was done trough usage of araylist.contains instead of having 11 conditions. All tests still work and checked cyclomatic complexity with lizard.


garydgregory:
This might satisfy some static code analysis but it's -1 for me since this PR now wastes memory allocating and building a List every single time. A bad trade off IMO.

To make this even worth, the PR also looses a code comment. Was this code change machine generated? That often happens with machine transforms.

What do others think?

kinow (commented):
@garydgregory agree. Imaging is a project that static analyzers and IDE's keep finding ways to improve, but that's normally better to leave the code as-is (to match specs, or as in this case, as it only satisfies the linter).

PR #280, opened 2023-02-20
Refactor BasicCParser::unescapeString()

snumlautoken:
BasicCParser::unescapeString() is currently one of the methods with the highest cyclometic complexities in the entire project. The purpose of the refactoring is to reduce complexity and to add consistency in regards to the switch-case.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/common/BasicCParser.java
+++ b/src/main/java/org/apache/commons/imaging/common/BasicCParser.java
@@ -342,6 +342,14 @@ public class BasicCParser {
         }
     }
 
+    /**
+    * Parses the hexadecimal-base escape-sequence found at index i of string.
+    * Helper-function for unescapeString()
+    * @param  i  the index of the escape-sequence in the string
+    * @param  stringBuilder the stringBuilder to append the escape-char to
+    * @param  string the string whose chars are parsed
+    * @return the new index i
+    */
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/imaging/common/BasicCParser.java
```diff
@@ -417,4 +341,103 @@ public static void unescapeString(final StringBuilder stringBuilder, final Strin
                     + "unterminated escape sequence found in string");
         }
     }
+
+    private static int appendHex(int i, final StringBuilder stringBuilder, final String string)
```
kinow, line 345: Could you add some javadoc to these methods too, please, @snumlautoken ? The names are self-explanatory, but might still be helpful to be explicit about it, and also mention the use of the parameters.

kinow (commented):
Hi @snumlautoken , thanks a lot for your PR.

The refactor seems to create a good separation of concerns with these new private methods :+1: 

I've enabled the GH Action workflow and left one comment. Let's wait for the workflow to finish to see the coverage (should ideally be the same or higher, although it could vary due to new lines for method names, etc).

Cheers
-Bruno

src/main/java/org/apache/commons/imaging/common/BasicCParser.java
snumlautoken: This has been added!

kinow:
Merged, thanks!

PR #283, opened 2023-03-07
[IMAGING-105] Add methods for editing EXIF-value for rotation for jpeg-files

snumlautoken:
This PR is for an issue whose 10th anniversary was a week ago, so might not be the most important issue.

Nevertheless, we feel as though, among metadata operations, rotation is probably the one most commonly used, and thus most deserving of a more streamlined process.
 
We also feel that our contribution could serve as a base for other streamlinings of exif-data manipulation, such as for GPS-data.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
+++ b/src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
@@ -1,0 +1,173 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.imaging.formats.jpeg.exif;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+
+import org.apache.commons.imaging.ImageReadException;
+import org.apache.commons.imaging.ImageWriteException;
+import org.apache.commons.imaging.Imaging;
+import org.apache.commons.imaging.common.bytesource.ByteSource;
+import org.apache.commons.imaging.common.bytesource.ByteSourceArray;
+import org.apache.commons.imaging.common.bytesource.ByteSourceFile;
+import org.apache.commons.imaging.formats.jpeg.JpegImageMetadata;
+import org.apache.commons.imaging.formats.tiff.TiffContents;
+import org.apache.commons.imaging.formats.tiff.TiffHeader;
+import org.apache.commons.imaging.formats.tiff.TiffImageMetadata;
+import org.apache.commons.imaging.formats.tiff.constants.TiffTagConstants;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputDirectory;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputField;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputSet;
+
+public class ExifOrientationRewriter {
+
+    public enum Orientation {
+        HORIZONTAL((short)1),
+        MIRROR_HORIZONTAL((short)2),
+        ROTATE_180((short)3),
+        MIRROR_VERTICAL((short)4),
+        MIRROR_HORIZONTAL_AND_ROTATE_270((short)5),
+        ROTATE_90((short)6),
+        MIRROR_HORIZONTAL_AND_ROTATE_90((short)7),
+        ROTATE_270((short)8);
+
+        private short val;
+
+        Orientation(short orVal) {
+            this.val = orVal;
+        }
+
+        public short getVal() {
+            return val;
+        }
+    }
+
+    private ByteSource fileSrc;
+
+    public ExifOrientationRewriter(File imageFile) {
+        fileSrc = new ByteSourceFile(imageFile);
+    }
+    public ExifOrientationRewriter(byte[] byteArray) {
+        fileSrc = new ByteSourceArray(byteArray);
+    }
+    public ExifOrientationRewriter(ByteSource byteSource) {
+        fileSrc = byteSource;
+    }
+
+    /***
+     * Get the orientation (enum) of the current image
+     * Returns horizontal by default
+     * @return Orientation enum
+     * @throws IOException
+     * @throws ImageReadException
+     * @throws ImageWriteException
+     */
+    public Orientation getExifOrientation() throws IOException, ImageReadException, ImageWriteException {
+
+        final JpegImageMetadata metadata = (JpegImageMetadata) Imaging.getMetadata(this.fileSrc.getAll());
+
+        if (metadata == null) {
+            return Orientation.HORIZONTAL;
+        }
+
+        final TiffImageMetadata exifMetadata = metadata.getExif();
+
+        if (exifMetadata == null) {
+            return Orientation.HORIZONTAL;
+        }
+
+        final TiffOutputSet outputSet = exifMetadata.getOutputSet();
+
+        TiffOutputDirectory tod = outputSet.getRootDirectory();
+        if (tod == null) {
+            return Orientation.HORIZONTAL;
+        }
+
+        TiffOutputField tof = tod.findField(TiffTagConstants.TIFF_TAG_ORIENTATION);
+        if (tof == null) {
+            return Orientation.HORIZONTAL;
+        }
+
+        short imageOrientationVal = (short) exifMetadata.getFieldValue(TiffTagConstants.TIFF_TAG_ORIENTATION);
+
+        for (Orientation orientation : Orientation.values()) {
+            if(orientation.getVal() == imageOrientationVal) {
+                return orientation;
+            }
+        }
+
+        return Orientation.HORIZONTAL;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


kinow (changes requested):
@snumlautoken there are over thirty commits in this PR. I think something went wrong when you created the branch or committed your changes??

snumlautoken:
@kinow 
We have been working as a group, and have thus been aiming to work in a very fractured manner, with many small commits. So no error on that front.
But a squash may have been appropriate...

snumlautoken:
Should we send in a new, squashed, PR?
Or just let it be?

snumlautoken:
Maybe not the ideal way to add co-authors, but this will hopefully work

kinow:
> Maybe not the ideal way to add co-authors, but this will hopefully work

If you would like to credit others, we can add it to the `src/changes/changes.xml`. That's where we record the entry for the issue (`IMAGING-105`). Alternatively, you can add multiple authors to a commit too - e.g. https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors (feel free to edit your commit and `push -f` to this branch as you did, until you are happy with the code :+1: )

snumlautoken:
> e.g. https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

That was the way I added them, so it should be fine as it is!

snumlautoken:
Re-requesting review as the change is still requested
Just making sure that a misunderstandiung has not occured

kinow:
No misunderstanding. Just busy, sorry. Thanks for checking though 👍

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+import org.apache.commons.imaging.formats.jpeg.JpegImageMetadata;
+import org.apache.commons.imaging.formats.tiff.TiffContents;
+import org.apache.commons.imaging.formats.tiff.TiffHeader;
+import org.apache.commons.imaging.formats.tiff.TiffImageMetadata;
+import org.apache.commons.imaging.formats.tiff.constants.TiffTagConstants;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputDirectory;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputField;
+import org.apache.commons.imaging.formats.tiff.write.TiffOutputSet;
+
+public class ExifOrientationRewriter {
```
kinow, line 42: Missing javadoc with `@since` tag (see other classes).

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public enum Orientation {
```
kinow, line 44: At least everything `public` must have a Javadoc :point_down:  (tests should be fine as we don't have that in other tests, unless you have time to add it too).

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+        fileSrc = byteSource;
+    }
+
+    /***
+     * Get the orientation (enum) of the current image
+     * Returns horizontal by default
+     * @return Orientation enum
+     * @throws IOException
+     * @throws ImageReadException
+     * @throws ImageWriteException
```
kinow, line 83: These also need a comment explaining the exceptions.

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        return Orientation.HORIZONTAL;
+    }
+
+    /**
+     * A method that sets a new value to the orientation field in the EXIF metadata of a JPEG file.
+     * @param orientation the value as a enum of the direction to set as the new EXIF orientation
+     *
```
kinow, line 125: Missing `@throws` here...

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * @return the ByteSource of the current file
+     */
+    public ByteSource getOutput() {
+        return fileSrc;
+    }
+
+    /**
+     *  Writes Bytesource to file with given path
```
kinow, line 161: Extra space before Writes, and s/Bytesource/ByteSource

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+     * @param path String of the path in which the file is written
+     * @throws IOException
+     */
+    public void getOutput(String path)
```
kinow, line 165: I think this is never tested?

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+    public ExifOrientationRewriter(byte[] byteArray) {
+        fileSrc = new ByteSourceArray(byteArray);
+    }
+    public ExifOrientationRewriter(ByteSource byteSource) {
+        fileSrc = byteSource;
+    }
+
+    /***
+     * Get the orientation (enum) of the current image
```
kinow, line 78: Needs a dot. The other sentence can come directly after this one, or leave it in the next line (or if you intended to have two lines, use HTML to add a break line, new paragraph, etc.).

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+     * @throws ImageReadException
+     * @throws ImageWriteException
+     */
+    public Orientation getExifOrientation() throws IOException, ImageReadException, ImageWriteException {
+
+        final JpegImageMetadata metadata = (JpegImageMetadata) Imaging.getMetadata(this.fileSrc.getAll());
+
+        if (metadata == null) {
+            return Orientation.HORIZONTAL;
+        }
```
kinow, line 91: Isn't that dangerous to return horizontal? Wouldn't it be better to return `null` instead? This way you have the three valid states, 1) data has horizontal orientation, 2) data has vertical orientation, or 3) no metadata about orientation.

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,173 @@
...
+     * A method that sets a new value to the orientation field in the EXIF metadata of a JPEG file.
+     * @param orientation the value as a enum of the direction to set as the new EXIF orientation
+     *
+     */
+    public void setExifOrientation(Orientation orientation) throws ImageWriteException, IOException, ImageReadException {
+
+        JpegImageMetadata metadata = (JpegImageMetadata) Imaging.getMetadata(this.fileSrc.getAll());
+
+        if (metadata == null) {
+            metadata = new JpegImageMetadata(null, new TiffImageMetadata(new TiffContents(new TiffHeader(ByteOrder.BIG_ENDIAN, 0, 0), new ArrayList<>(), new ArrayList<>())));
```
kinow, line 132: I think `ExifRewriter` allows big and little endian. But it looks like `ExifOrientationRewriter` uses big endian if no metadata found? Wouldn't be better if we had the same behavior as in `ExifRewriter`? Letting the user choose the endianess.

kinow: s/new ArrayList()/Collections.emptyList()

That should create that empty `EmptyList` instead of an `ArrayList`, and I think further down the pipe the `new ArrayList` will become an unmodifiable collection, so that should be fine.

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
kinow, line 1: Did you try to design the EXIF rotation as part of `ExifRewriter`, @snumlautoken ? What was the rationale for creating a new class instead? I guess that could mean that for other metadata rewriting/editing we would need other `Exif$MetadataRewriter`? Ditto for other metadata/formats? That sounds like it could become problematic in the long run, I think? WDYT?

kinow (changes requested):
The code is in excellent condition, @snumlautoken ! I added some comments about Javadoc and other things related to the current code style. Also some comments/questions about some design choices for the `ExifOrientationRewriter`. Sorry for the delay :wave: 

-Bruno

StefanOltmann:
@snumlautoken I share your opinion about the importance about the orientation flag. That’s why I recently wrote a logic dedicated to finding and updating that flag.

You may want to rewrite https://github.com/Ashampoo/kim/blob/main/src/commonMain/kotlin/com/ashampoo/kim/format/jpeg/JpegOrientationOffsetFinder.kt in Java and contribute this. :)

kinow:
Rebased.

kinow:
Added a short WIP PR updating the internal API as that changed a bit since the PR was created. Will ping @garydgregory in a commit as I couldn't find a replacement for the old `ByteSource.getAll` that's used here.

src/main/java/org/apache/commons/imaging/formats/jpeg/exif/ExifOrientationRewriter.java
```diff
@@ -0,0 +1,168 @@
...
+        return Orientation.HORIZONTAL;
+    }
+
+    /**
+     * A method that sets a new value to the orientation field in the EXIF metadata of a JPEG file.
+     * @param orientation the value as a enum of the direction to set as the new EXIF orientation
+     */
+    public void setExifOrientation(Orientation orientation) throws IOException {
+
+        JpegImageMetadata metadata = (JpegImageMetadata) Imaging.getMetadata(this.fileSrc.getAll());
```
kinow, line 124: `fileSrc` is a `ByteSource`, but getAll was removed: https://github.com/apache/commons-imaging/commit/f0b55d0a

kinow:
Hi @nimlapong , are you still willing to contribute this to Imaging? I noticed you deleted the repository, but I had saved it in the list of issues to fix someday. Cheers

PR #291, opened 2023-04-06
refactored classes IcnsImageParser, PhotometricInterpreterLogLuv, ByteSourceArray, PcxWriter, pivotProcessor, ImageParser, PsdImageParser, PngImageParser for code smells. 

NaveenKunapaneni:
IcnsImageParser -> Refactored a complex method(writeImage) to multiple smaller components (getImageType(src), writeHeader(bos, imageType), writeImageData(bos, src, imageType), writeMaskData(bos, src, imageType),
 
PhotometricInterpreterLogLuv ->Introduced a new variable expected_bit_per_sample, 

ByteSourceArray -> Deomposed a complex contional if statement in getBlock method, 

PcxWriter -> Moved a method writePixel32 from PcxWriter to RleWriter class, 

pivotProcessor -> Extracted methods which calculates pivot related values and pushed them to a new class , 

ImageParser -> pushed down a method form the parent class (ImageParser) to child ( PsdImageParser PngImageParser) classes.

 I am currently studying on smells and various techniques to refactor smells. This is a small effort on getting an under standing code smells and refactoring them.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="MavenProjectsManager">
+    <option name="originalFiles">
+      <list>
+        <option value="$PROJECT_DIR$/pom.xml" />
+      </list>
+    </option>
+  </component>
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_X" default="true" project-jdk-name="openjdk-20" project-jdk-type="JavaSDK" />
+</project>
--- a/src/main/java/org/apache/commons/imaging/color/ColorConversions.java
+++ b/src/main/java/org/apache/commons/imaging/color/ColorConversions.java
@@ -29,12 +29,6 @@ public final class ColorConversions {
     /** see: https://en.wikipedia.org/wiki/CIELAB_color_space#From_CIEXYZ_to_CIELAB[10] */
     private static final double REF_Z = 108.883;
 
-    /** see: https://en.wikipedia.org/wiki/CIELAB_color_space#From_CIEXYZ_to_CIELAB[10] */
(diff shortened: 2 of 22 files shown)


.idea/misc.xml
kinow, line 1: Ide files are not included with imaging.

src/main/java/org/apache/commons/imaging/color/ColorConversions.java
```diff
@@ -49,9 +43,9 @@ public static ColorCieLab convertXYZtoCIELab(final double X, final double Y, fin
         double var_Z = Z / REF_Z; // REF_Z = 108.883
 
         // Pivot XÝZ:
-        var_X = pivotXYZ(var_X);
-        var_Y = pivotXYZ(var_Y);
-        var_Z = pivotXYZ(var_Z);
+        var_X = pivotProcessor.pivotXYZ(var_X);
+        var_Y = pivotProcessor.pivotXYZ(var_Y);
+        var_Z = pivotProcessor.pivotXYZ(var_Z);
```
kinow, line 48: Variables in Imaging may have weird names but these many times match what is described on papers and specifications. Refactorings like these are not really helpful.

kinow (changes requested):
Hi. Had a quick look but immediately noticed IDE files added to source and changes to code like remove unused variables, comments to with links to docs, and renaming variables.

These are not really helpful since the code in Imaging normally matches code in other implementations, or pseudocode from papers and spwcifications. Renaming x1 to something more meaningful is normally a good pracrice. But in the case of Imaging it is more important to quickly know what is the x1 mentioned in some spec or soc. We can still improve this with comments (these are always welcome).  Will have a look later to see if there are other changes that can be added, but these changes (ide and IDE/linter batch renaming and removal) would need to be dropped.

Thanks!!!

garydgregory:
I agree with @kinow

garydgregory:
Closing: No reply from the author.

PR #294, opened 2023-05-23
IMAGING-352-palettefactory: IMAGING-352: allow supplying a custom Pal…

jvdvegt:
This fixes https://issues.apache.org/jira/browse/IMAGING-352 : it allows the user to supply their own version of the PaletteFactory class.

Example usage: https://github.com/jvdvegt/WuQuant/blob/77ccea6a28193ab1261528c555699a9c365b980a/src/test/java/nl/jvdvegt/WuQuantTest.java#L44

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
+++ b/src/main/java/org/apache/commons/imaging/formats/png/PngWriter.java
@@ -32,7 +32,7 @@ import org.apache.commons.imaging.internal.Debug;
 import org.apache.commons.imaging.palette.Palette;
 import org.apache.commons.imaging.palette.PaletteFactory;
 
-class PngWriter {
+public class PngWriter {
 
     /*
      1. IHDR: image header, which is the first chunk in a PNG datastream.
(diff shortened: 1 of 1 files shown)


garydgregory:
@jvdvegt 
Thank you for your PR. It is missing unit tests for the new public APIs.

jvdvegt:
@garydgregory Thanks, I'm on it.

While writing a test, I noticed a potential bug in PngImageParser: it only sets 'usesPalette' when there are *more than one* palette (see https://github.com/apache/commons-imaging/blob/a6ccaa76fdf1627a402eb7a2b965e457a7733968/src/main/java/org/apache/commons/imaging/formats/png/PngImageParser.java#L500 ). That contradicts the check in that same file in 'getBufferedImage': https://github.com/apache/commons-imaging/blob/a6ccaa76fdf1627a402eb7a2b965e457a7733968/src/main/java/org/apache/commons/imaging/formats/png/PngImageParser.java#L180
I guess a PNG file can at most contain one palette, so the check on line 500 should be for 'isEmpty'.

Shall I fix that in my PR as well, or does it need a separate ticket?

garydgregory:
Hello @jvdvegt 
It might be best in separate PRs so either can be reverted if the need arises.

PR #301, opened 2023-07-08
[IMAGING-355] Large animated GIF takes too much heap memory in getMetadata

andrewmkhoury:
IMAGING-355 - Large animated GIF takes too much heap memory in getMetadata

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
@@ -407,7 +407,7 @@ public class GifImageParser extends ImageParser<GifImagingParameters> implements
     @Override
     public ImageInfo getImageInfo(final ByteSource byteSource, final GifImagingParameters params)
             throws ImagingException, IOException {
-        final GifImageContents blocks = readFile(byteSource, false);
+        final GifImageContents blocks = readFile(byteSource, true);
 
         final GifHeaderInfo bhi = blocks.gifHeaderInfo;
         if (bhi == null) {
@@ -482,7 +482,7 @@ public class GifImageParser extends ImageParser<GifImagingParameters> implements
     @Override
     public ImageMetadata getMetadata(final ByteSource byteSource, final GifImagingParameters params)
             throws ImagingException, IOException {
-        final GifImageContents imageContents = readFile(byteSource, false);
+        final GifImageContents imageContents = readFile(byteSource, true);
 
         final GifHeaderInfo bhi = imageContents.gifHeaderInfo;
         if (bhi == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -407,7 +407,7 @@ public byte[] getIccProfileBytes(final ByteSource byteSource, final GifImagingPa
     @Override
     public ImageInfo getImageInfo(final ByteSource byteSource, final GifImagingParameters params)
             throws ImagingException, IOException {
-        final GifImageContents blocks = readFile(byteSource, false);
+        final GifImageContents blocks = readFile(byteSource, true);
```
kinow, line 410: Maybe this `true`/`false` (`stopBeforeImageData`) should instead be retrieved from the `GitImagingParameters` object?

garydgregory: That's a good point!

I wonder if this data is hard-coded to be skipped it will be unavailable for other API calls once the object is constructed. That would be bad. But if it can be paramerized, then the client call.site can decide.

garydgregory: **Note**: All current callers of `org.apache.commons.imaging.formats.gif.GifImageParser.readFile(ByteSource, boolean)` pass false for `stopBeforeImageData`.

Would it make for all call sites to pick up this value from `GifImagingParameters`? Or, would this cause some use-cases to break. We will need to make sure to unit test whatever code we change or add.

andrewmkhoury: `stopReadingBeforeImageData` or `skipImageData` works.

andrewmkhoury:
@garydgregory and @kinow I implemented the changes per your request.  However, I am not sure how much value passing stopReadinngBeforeImageData adds.  getImageInfo already has a unit test and I added a test for getMetadata.
* Add stopReadingBeforeImageData to GifImageParameters
* Add unit test for IMAGING-355

src/main/java/org/apache/commons/imaging/formats/gif/GifImageParser.java
```diff
@@ -407,7 +407,11 @@ public byte[] getIccProfileBytes(final ByteSource byteSource, final GifImagingPa
     @Override
     public ImageInfo getImageInfo(final ByteSource byteSource, final GifImagingParameters params)
             throws ImagingException, IOException {
-        final GifImageContents blocks = readFile(byteSource, false);
+        boolean stopReadingBeforeImageData = true;
```
garydgregory, line 410: You've changed the default behavior here which makes the parameter a forced usage, which might be a surprised.

PR #303, opened 2023-07-15
IMAGING-322: ImageFormat uses a list for extensions instead of array

orionlibs:
[IMAGING-322](https://issues.apache.org/jira/browse/IMAGING-322): ImageFormat uses a list for extensions instead of array

### DIFF
--- a/src/main/java/org/apache/commons/imaging/ImageFormats.java
+++ b/src/main/java/org/apache/commons/imaging/ImageFormats.java
@@ -16,6 +16,11 @@
  */
 package org.apache.commons.imaging;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 /**
  * Enum of known image formats.
  */
@@ -43,24 +48,29 @@ public enum ImageFormats implements ImageFormat {
     XBM("xbm"),
     XPM("xpm");
 
-    private final String[] extensions;
+    private final List<String> extensions;
 
-    ImageFormats(final String ...extensions) {
-        this.extensions = extensions;
+    ImageFormats(final String... extensions) {
+        this.extensions = Arrays.stream(extensions).collect(Collectors.toList());
     }
 
     @Override
     public String getDefaultExtension() {
-        return this.extensions != null ? this.extensions[0] : null;
+        return this.extensions != null ? this.extensions.get(0) : null;
     }
 
     @Override
-    public String[] getExtensions() {
-        return this.extensions.clone();
+    public List<String> getExtensions() {
+        return new ArrayList<>(this.extensions);
     }
 
     @Override
     public String getName() {
         return name();
+    }
+
+
+    public static List<ImageFormats> valuesAsList() {
+        return Arrays.stream(values()).collect(Collectors.toList());
     }
 }
--- a/src/main/java/org/apache/commons/imaging/formats/pnm/PnmImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/pnm/PnmImageParser.java
@@ -26,6 +26,7 @@ import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.StringTokenizer;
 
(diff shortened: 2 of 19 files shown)


StefanOltmann:
That makes sense. In my fork I refactored it to a `Set`, because there is no order needed.

StefanOltmann:
That makes sense. In my fork I use a `Set`, because there is no order needed.

https://github.com/Ashampoo/kim/blob/2d437c4d6b54cf64450add84600a7af370fae36a/src/commonMain/kotlin/com/ashampoo/kim/model/ImageFormat.kt#L28C3-L28C3

orionlibs:
@StefanOltmann the ticket said List, tha tis why I did it with a list. Should I change it to Set?

StefanOltmann:
That’s for the reporter to decide, but I wanted just to add my two cents here. 😄

In general if you don’t need order and there are no duplicates Set should be used.

garydgregory:
@gwlucastrig @kinow Any thoughts on array vs. set vs. list?

orionlibs:
> @gwlucastrig @kinow Any thoughts on array vs. set vs. list?

it is true that there are no duplicate image extensions. Performance-wise:
1--an ArrayList, under the hood, uses a dynamically resizing array to store its elements. So, the memory consumption would be the memory taken by each extension plus some overhead for the array itself. If the array is not full, there can be some wasted memory.
2--a HashSet uses a HashMap under the hood. I think a HashSet would take more memory than an ArrayList as it needs to maintain the additional data structure (HashMap) with an extra constant value for every entry. The iterator of a HashSet is slower than an ArrayList because it needs to find the next non-null bucket every time.

If that is true then the current implementation may be enough.

kinow:
>@gwlucastrig @kinow Any thoughts on array vs. set vs. list?

Sorry, a bit out of the loop. I think Imaging adopts arrays in some places for performance, but there could be some other reason too (e.g. directly translation from C/Python without thinking too hard about lists vs. arrays, or simply copied what was already done in other parts of the code).

IMHO, consistency is also important. So if lists are used here, users would probably expect to find the extensions as lists in other parts, and maybe similar concepts in Imaging. But if there are parsers/utilities/etc still using arrays… then I'd have a wider look and confirm it really makes sense and also create an issue to plan & discuss (also in the ml list) switching.

src/main/java/org/apache/commons/imaging/ImageFormats.java
```diff
@@ -43,24 +48,29 @@ public enum ImageFormats implements ImageFormat {
     XBM("xbm"),
     XPM("xpm");
 
-    private final String[] extensions;
+    private final List<String> extensions;
 
-    ImageFormats(final String ...extensions) {
-        this.extensions = extensions;
+    ImageFormats(final String... extensions) {
+        this.extensions = Arrays.stream(extensions).collect(Collectors.toList());
```
Marcono1234, line 54: Why not `java.util.Arrays.asList(...)`? `Collectors.toList()` makes not guarantees about the created list, see [documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#toList()):
> There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned

If the intention here is to create a copy of the array, you could also do `Arrays.asList(extensions.clone())`; that would make the intention clearer then.
(But I don't think there is a need to copy the array here.)

src/main/java/org/apache/commons/imaging/ImageFormats.java
```diff
@@ -43,24 +48,29 @@ public enum ImageFormats implements ImageFormat {
...
     }
 
     @Override
     public String getName() {
         return name();
     }
+
+
+    public static List<ImageFormats> valuesAsList() {
+        return Arrays.stream(values()).collect(Collectors.toList());
```
Marcono1234, line 74: `Arrays.asList(...)` instead?

And is this new methods even used? If not maybe it should be omitted because it tries to work around a flaw with the standard Java enum methods (related to [JDK-8073381](https://bugs.openjdk.org/browse/JDK-8073381)); and fixing this in a library might not be the right place?

src/main/java/org/apache/commons/imaging/ImageFormats.java
```diff
@@ -43,24 +48,29 @@ public enum ImageFormats implements ImageFormat {
...
     public String getDefaultExtension() {
-        return this.extensions != null ? this.extensions[0] : null;
+        return this.extensions != null ? this.extensions.get(0) : null;
     }
 
     @Override
-    public String[] getExtensions() {
-        return this.extensions.clone();
+    public List<String> getExtensions() {
+        return new ArrayList<>(this.extensions);
```
Marcono1234, line 64: What about returning an unmodifiable list here (respectively directly making the list unmodifiable in the `ImageFormats` constructor)? It appears multiple parsers store a single instance of the result of this and then expose it through their `getAcceptedExtensions()` method, so this could lead to accidental modification.

src/main/java/org/apache/commons/imaging/formats/pnm/PnmImageParser.java
```diff
@@ -42,13 +43,13 @@
...
 
     private static final String DEFAULT_EXTENSION = ImageFormats.PNM.getDefaultExtension();
-    private static final String[] ACCEPTED_EXTENSIONS = {
-            ImageFormats.PAM.getDefaultExtension(),
-            ImageFormats.PBM.getDefaultExtension(),
-            ImageFormats.PGM.getDefaultExtension(),
-            ImageFormats.PNM.getDefaultExtension(),
-            ImageFormats.PPM.getDefaultExtension()
-    };
+    private static final List<String> ACCEPTED_EXTENSIONS = Arrays.asList(
```
Marcono1234, line 46: Should this additionally wrap the list in `Collections.unmodifiableList`? Because while the list returned by `Arrays.asList` does not support changing the size, it does still support replacing elements.

Marcono1234 (commented):
Hopefully my review comments below are helpful; however, feel free to consider them only as suggestions since I am not a member of this project.

(Sorry if these review comments are too intrusive, in that case please let me know.)

PR #304, opened 2023-07-20
Make checkForSubImage null-safe

mercuryLuz:
Make checkForSubImage null-safe for TiffImagingParameters params.
Especially as params gets set as null in TiffImageParser.java:183 getAllBufferedImages()

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
@@ -75,7 +75,7 @@ public class TiffImageParser extends ImageParser<TiffImagingParameters> implemen
         // dimensions of the image that is being read.  This method
         // returns the sub-image specification, if any, and leaves
         // further tests to the calling module.
-        if (params.isSubImageSet()) {
+        if (params != null && params.isSubImageSet()) {
             final int ix0 = params.getSubImageX();
             final int iy0 = params.getSubImageY();
             final int iwidth = params.getSubImageWidth();


src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
```diff
@@ -75,7 +75,7 @@ private Rectangle checkForSubImage(
         // dimensions of the image that is being read.  This method
         // returns the sub-image specification, if any, and leaves
         // further tests to the calling module.
-        if (params.isSubImageSet()) {
+        if (params != null && params.isSubImageSet()) {
```
kinow, line 78: Hadn't time to review properly, but it would be useful to check if the method that calls this private method couldn't be updated instead, to avoid a null object reaching here. Thanks

mercuryLuz: Thanks for the quick response!
There are multiple calling methods and IMHO it makes the most sense within this method, so that any consumers can just use checkForSubImage(params) to find out whether there's any subImage data to be extracted no matter the given params.

garydgregory: The only thing missing here is a unit test that shows why this is a problem.

Ditscheridou:
I ran into this today and was just on he way to create a ticket. Any chance this can be merged?

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
Ditscheridou: This is an actual problem. If you want to read a multipage tiff with:

`TiffImageParser().getAllBufferedImages(File(path))`

you will run into a npe.

garydgregory:
> I ran into this today and was just on he way to create a ticket. Any chance this can be merged?

Hello @mercuryLuz 
Do you have an image that reproduces the issue? Would please create a PR with a unit test?

Ditscheridou:
Yeah I can provide that later today. Will open a new PR for that

src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
garydgregory: @Ditscheridou, @mercuryLuz 
This PR still needs a test.

garydgregory:
> Yeah I can provide that later today. Will open a new PR for that

Hello @Ditscheridou 

Please follow up or close this PR. TY!

garydgregory:
Closing: In git master now. You are credited in changes.xml.

PR #306, opened 2023-07-23
[IMAGING-359] Fix test resource file paths not being constructed properly

Marcono1234:
Using `URL.getFile()` is faulty because its result is URL encoded. This caused test failures when running in a directory with spaces in its file path.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/formats/bmp/BmpImageParserTest.java
+++ b/src/test/java/org/apache/commons/imaging/formats/bmp/BmpImageParserTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.imaging.formats.bmp;
 
+import static org.apache.commons.imaging.test.TestResources.fileResource;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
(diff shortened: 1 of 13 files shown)


garydgregory:
Where are the test failures? The builds are all green here.

Marcono1234:
I meant caused test failures _previously_. Simply clone the current state of `master` into a directory with spaces in its name and try to run the tests; you will see a lot of `NoSuchFileException`s being thrown, with the file path containing the space encoded as `%20`.

src/test/java/org/apache/commons/imaging/test/TestResources.java
```diff
@@ -0,0 +1,59 @@
...
+     * Gets a {@link File} for a {@linkplain Class#getResource(String) resource path}.
+     */
+    public static File fileResource(String path) {
+        return new File(resourceUri(path));
+    }
+
+    /**
+     * Gets a {@link Path} for a {@linkplain Class#getResource(String) resource path}.
+     */
+    public static Path pathResource(String path) {
```
Marcono1234, line 56: I am not completely sure about those names; at least `fileResource` sounds a bit like it also verifies the resource is an existing file (and not a directory), which it doesn't.

Maybe `resourceFile` and `resourcePath` would be better (or some other name)? What do you think?

src/test/java/org/apache/commons/imaging/formats/bmp/BmpImageParserTest.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.imaging.formats.bmp;
 
+import static org.apache.commons.imaging.test.TestResources.fileResource;
```
garydgregory, line 19: I think we should only use static imports for JUnit.

garydgregory (changes requested):
See comment.

src/test/java/org/apache/commons/imaging/formats/bmp/BmpImageParserTest.java
Marcono1234: Have changed it to regular imports

garydgregory:
https://issues.apache.org/jira/browse/IMAGING-359

PR #307, opened 2023-07-23
Use static imports for JUnit Assertions methods

Marcono1234:
Currently `static` imports aren't used consistently for JUnit's `org.junit.jupiter.api.Assertions` methods.

### DIFF
--- a/src/test/java/org/apache/commons/imaging/bytesource/ByteSourceTest.java
+++ b/src/test/java/org/apache/commons/imaging/bytesource/ByteSourceTest.java
@@ -17,6 +17,7 @@
 package org.apache.commons.imaging.bytesource;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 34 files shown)


garydgregory (changes requested):
@Marcono1234 
Thank you for your PR.
Don't change the order of import statements please. It adds noise to the PR and makes it take longer to review.

Marcono1234:
> Don't change the order of import statements please

I think all the changes in this pull request follow the existing import ordering, though please point me to a specific location where that is not true in case I overlooked something.

Since this pull request changes the test code to consistently only use `static` imports for JUnit's `Assertions` methods, it has to replace the regular `import` statements with `import static`, which therefore also has to change the "order" (actually not really, because the imports themselves change) to keep following the existing import ordering, that is, `static` imports first.

garydgregory:
@Marcono1234 you're right! I'll review tonight.

PR #311, opened 2023-08-17
Issue-360: Add more TIFF compression entries to ImageInfo

gwlucastrig:
This pull request adds more information about TIFF data compression variations to the ImageInfo available for TIFF files.

**Detail from Issue-360**

The ImageInfo for TIFF is missing information related to three TIFF compression options, two of which are quite commonly used. I propose to modify the TiffImageParser class to populate these values.
 
```
TIFF_COMPRESSION_DEFLATE_PKZIP 
TIFF_COMPRESSION_DEFLATE_ADOBE 
```

Additionally, the TIFF format has an obsolete variation on JPEG that was abandoned in the 1990's.  Commons Imaging does not currently support TIFF JPEG compression, though many of the pieces needed are already in its code base.  I propose to do the following:

The current code has a class called TiffConstants that includes the definition for
 
`    TIFF_COMPRESSION_JPEG = 6; `

Code 6 is the obsolete JPEG variation.  Code 7 is the format currently used in many TIFF utilities. So I propose to modify the constants and as follows
 
```
TIFF_COMPRESSION_JPEG_OBSOLETE = 6
TIFF_COMPRESSION_JPEG = 7
```

### DIFF
--- a/src/main/java/org/apache/commons/imaging/ImageInfo.java
+++ b/src/main/java/org/apache/commons/imaging/ImageInfo.java
@@ -57,6 +57,7 @@ public class ImageInfo {
         NONE("None"),
         LZW("LZW"),
         PACKBITS("PackBits"),
+        JPEG_TIFF_OBSOLETE("JPEG Obsolete (TIFF only)"),
         JPEG("JPEG"),
         RLE("RLE: Run-Length Encoding"),
         ADAPTIVE_RLE("Adaptive RLE"),
@@ -64,7 +65,8 @@ public class ImageInfo {
         PNG_FILTER("PNG Filter"),
         CCITT_GROUP_3("CCITT Group 3 1-Dimensional Modified Huffman run-length encoding."),
         CCITT_GROUP_4("CCITT Group 4"),
-        CCITT_1D("CCITT 1D");
+        CCITT_1D("CCITT 1D"),
+        DEFLATE("DEFLATE (ZIP)");
 
         private final String description;
 
--- a/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageParser.java
@@ -49,6 +49,9 @@ import org.apache.commons.imaging.common.ImageMetadata;
 import org.apache.commons.imaging.common.XmpEmbeddable;
 import org.apache.commons.imaging.common.XmpImagingParameters;
 import org.apache.commons.imaging.formats.tiff.TiffDirectory.ImageDataElement;
+import static org.apache.commons.imaging.formats.tiff.constants.TiffConstants.TIFF_COMPRESSION_DEFLATE_ADOBE;
+import static org.apache.commons.imaging.formats.tiff.constants.TiffConstants.TIFF_COMPRESSION_DEFLATE_PKZIP;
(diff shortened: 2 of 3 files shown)


garydgregory:
I don't think these names changes are great: what happens when 7 is obsolete in a year or two? Is 6 and 7 known version numbers? If so then maybe the constant names should contain the RFC or specification version.

gwlucastrig:
In answer to your question, 6 and 7 are integer codes that are embedded in a TIFF file to indicate which kind of data compression it uses.  There are a family of static variables defined in TiffConstants that define the codes for a compression. In a more modern implementation, we would probably opt to define the compression option using an enumeration, but the original authors wrote the TIFF code was written a long time ago, before Java enumerations really caught on.  Also, a constant for JPEG=7 isn't even defined in TiffConstants, only 6. Which is ironic since 6 was already obsolete in the 90's and option 7 was formally described in 2002 (see TIFF 6.0, Part 2: TIFF Extensions). 

I selected the TIFF constants in my pull request based on the idea of making as few changes to the code base as possible. I am currently implementing support for JPEG (see IMAGING-194, IMAGING-240) and wanted to pave the way for the code change coming in the future.  Since there was no constant for JPEG in TiffConstants, I needed to add one.  I wanted to stay within the style established by the original authors.  Also, the TIFF format is pretty settled (again, JPEG was specified in 2002), and I don't think there will be much change in the future in terms of new RFC's etc. 

Again, the number 7 is not a "version" per se.  It's a code embedded into TIFF files to indicate what kind of data compression they use.  The number 8, for example, indicates the DEFLATE (ZIP) format.  The number 6 is the old LZW compression.  The next one after 8 is 32773 for PACKBITS.    So there is no sequential scheme for numbering things.  Formats like JPEG 2002 were never adopted by the TIFF community.

gwlucastrig:
I take it you're not going to accept this PR.  So I want to give you a "heads up".  I am working on enabling Commons Imaging to read TIFF files that contain JPEG-formatted imagery.  When the time comes to submit it, I am going to have to do something to specify a constant for compression code 7.  I wanted to let you know that's coming so it doesn't look like I'm trying to "sneak something into the code base."

garydgregory:
@gwlucastrig 
I assume you're talking to me about accepting THIS PR. I've reviewed it again, so I can't say.

gwlucastrig:
Your assumption is correct.  I am going to submit the JPEG changes in a future PR.  That PR could build on this one (PR #311) if it were accepted.  But, if this PR were not accepted, the future PR would require a specification for the JPEG compression code value of 7.  If this PR is not accepted, the future PR would need to do _something_  to provide that constant.  I just wanted to make sure that when the time came, and the future PR was submitted, it would not appear like I was doing something sneaky to get my preferred changes into the code base.

The JPEG changes are a lot of work and it will be weeks before I turn them in. So this discussion does not merit an immediate response if you are busy.

garydgregory:
Hello @gwlucastrig 
Thank you for your patience.
OK, seems fine. I can merge this now or should I wait for #311?

gwlucastrig:
I believe that these changes are "correct".  But now that I've learned a bit more about how the Apache PR process works, I'm concerned.  You might now want to take this PR because I do not think it will improve the coverage.  It may even degrade it by a very small percentage (since I added an exception for which there is no test case).

If you are comfortable with this change, I'd say go ahead and pull it in.    Otherwise, I will incorporate these changes when I turn in my changes for the TIFF-JPEG format, Jira Issues Imaging-194 and Imaging-240

PR #339, opened 2023-11-27
Code Refactoring

ashishnagpal2498:
Worked on implementation smells.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/palette/Dithering.java
+++ b/src/main/java/org/apache/commons/imaging/palette/Dithering.java
@@ -24,39 +24,35 @@ import org.apache.commons.imaging.ImagingException;
  * Dithering algorithms to use when quantizing an image to palette form.
  */
 public final class Dithering {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/imaging/palette/Dithering.java
```diff
@@ -76,42 +72,33 @@ public static void applyFloydSteinbergDithering(final BufferedImage image, final
...
-                    if (x - 1 >= 0) {
-                        update = adjustPixel(image.getRGB(x - 1, y + 1), errA, errR, errG, errB, 3);
-                        image.setRGB(x - 1, y + 1, update);
-                    }
-
-                }
             }
         }
     }
-
```
garydgregory, line 114: -1: You've now made the constructor public, which we do not want since this would allow users to create instances. I added a comment here.

garydgregory: @ashishnagpal2498 
It's quite a waste of time to review moving code around, please don't do it.

garydgregory:
Closing: No reply from author.

PR #342, opened 2023-12-12
Maven wrapper

asomov:
- fix a test
- ignore IDEA files
- use [Maven wrapper ](https://maven.apache.org/wrapper/)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -45,7 +45,7 @@ jobs:
         distribution: 'temurin'
         java-version: ${{ matrix.java }}
     - name: Build with Maven
-      run: mvn --show-version --batch-mode --no-transfer-progress test jacoco:report
+      run: ./mvnw --show-version --batch-mode --no-transfer-progress test jacoco:report
 
     - name: Upload coverage to Codecov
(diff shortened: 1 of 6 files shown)


garydgregory:
No Maven wrapper please.

asomov:
> No Maven wrapper please.

Can you please clarify the statement ?
Maven wrapper is the only way to use the very same Maven for all the developers and CI. It is the way to make the repetitive builds

asomov:
@garydgregory what are disadvantages to have Maven wrapper configured ?

asomov:
> As to the renaming of the TIFF file, it is unnecessary, as the only test using it converts the name to lower-case.

@sebbASF the test fails for me locally. I wonder why it does not for you.

asomov:
> Regardless of the merits of the wrapper, I don't see how it applies to the Github CI builds.
> 
> Also, it would make the GH workflow files harder to maintain, as the command would different on Windows.

@sebbASF how can see the version of Maven used for all the builds in CI ?

asomov:
@sebbASF thank for the clarification. We use docker, which means that from local development and testing to any remote deployment the code is under Linux.
I can understand that Windows is supported together with Linux, but I do not get how Windows is supported instead of Linux. 
I am afraid to stay on a safe side, I would rather choose another library.

PR #368, opened 2024-02-28
IMAGING-373: Throw exception on invalid bitmap offset

Draczech:
Fix of an issue described in https://issues.apache.org/jira/projects/IMAGING/issues/IMAGING-373

In case there are less bytes available in loaded image file than **bitmapPixelsOffset** says, an exception is thrown. New test case created with a broken ICO file to show how the problem is fixed.


src/test/java/org/apache/commons/imaging/formats/ico/IcoImageParserTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.commons.imaging.test.TestResources;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+
+public class IcoImageParserTest {
+
+    /**
+     * For https://issues.apache.org/jira/browse/IMAGING-373.
+     * <p> There is a problem with loading bitmap stored in given ICO file, so the exception is originally thrown by BmpImageParser.
```
garydgregory, line 31: Close HTML tags. No need for a space after `<p>`.

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
```diff
@@ -569,7 +569,7 @@ private BmpImageContents readImageContents(final InputStream is, final FormatCom
             debugNumber("expectedDataOffset", expectedDataOffset, 4);
         }
         final int extraBytes = bhi.bitmapDataOffset - expectedDataOffset;
-        if (extraBytes < 0) {
+        if ((extraBytes < 0) || (extraBytes > bhi.fileSize)) {
```
garydgregory, line 572: No need for extra parens.

garydgregory (changes requested):
Hello @Draczech 
Thank you for your PR.
Please see my comments.

src/test/java/org/apache/commons/imaging/formats/ico/IcoImageParserTest.java
Draczech: Fixed.

I haven't noticed any plugin like https://github.com/diffplug/spotless in the project, I might just missed it. Apparently I'm relying too much on such tools :)

src/main/java/org/apache/commons/imaging/formats/bmp/BmpImageParser.java
Draczech: Fixed

garydgregory:
@Draczech 
TY !

PR #451, opened 2024-10-17
Add support of GPSHPositioningError

Stefal:
Hi!

I'm not a Java developer at all. I've done my best to understand the code and add the support of the GPSHPositioningError tag. This tag seems available since [Exif 2.3](https://home.jeita.or.jp/tsc/std-pdf/CP3451C.pdf)

```
GPSHPositioningError
This tag indicates horizontal positioning errors in meters.
Tag = 31 (1F.H)
Type = RATIONAL
Count = 1
Default = None

Note When the tag Type is ASCII, it shall be terminated with NULL.
It shall be noted carefully that since the value count includes the terminator NULL, the total
count is the number of data+1. For example, GPSLatitudeRef cannot have any values other
than Type ASCII 'N' or 'S'; but because the terminator NULL is added, the value of N is 2.
```


garydgregory:
Hello @Stefal 

Thank you for your PR.

New public and protected elements should have a Javadoc comment that includes a since tag.

This PR needs a unit test to prove the change does what you think it does.

It's OK to include an image with the test as long as it is yours, or if not, its license is Apache 2.0 or compatible with the Apache license.

garydgregory:
Hello @Stefal 

Thank you for your updates.

Please run `mvn` before you push to run and fix all build checks.

I'm not sure which exact version of Exif we support but that would be good to document.

Stefal:
Hello @garydgregory 

I've fixed the 2 CheckStyle errors Maven reported.

Stefal:
Is there anything more I need to do to get this pull request merged?

>I'm not sure which exact version of Exif we support but that would be good to document.

if https://github.com/apache/commons-imaging/pull/49 was merged we could say it's v2.32 😄

garydgregory:
@Stefal 
I'll review later today or tomorrow. 
Ty!

Stefal:
Thank you @garydgregory

garydgregory:
@Stefal 
Yw!

PR #477, opened 2025-01-21
Optionally force UTF-8 encoding for IPTC records

thstieler:
This pull request adds the option to enforce UTF-8 encoding for writing IPTC records to jpeg files.

This can improve the reliability for correct parsing of IPTC records written by commons-imaging, because some libraries/applications use UTF-8 as default charset when reading IPTC records.


garydgregory:
@thstieler 

It looks like you did not follow the PR template, specfically:
- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.

thstieler:
@garydgregory You're right, unfortunately I forgot to remove trailing whitespace in one line, now mvn completes successfully!

thstieler:
@garydgregory Thank you very much, that's great :-)!

PR #507, opened 2025-05-14
Update TiffImageMetadata.java

clagenna:
Due to some HEIC file converters to JPG longitude and latitude strings gets converted from UNICODE  with wrong UTF-8 coding. e.g. North "N" have hex bytes hex:4e ef bf bd For that reason latitudeRef.trim().equalsIgnoreCase("n") fails. My suggestion  is to use substring(0, 1) instead of trim()

Thanks for your contribution to [Apache Commons](https://commons.apache.org/)! Your help is appreciated!

Before you push a pull request, review this list:

- [ ] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageMetadata.java
+++ b/src/main/java/org/apache/commons/imaging/formats/tiff/TiffImageMetadata.java
@@ -173,11 +173,15 @@ public class TiffImageMetadata extends GenericImageMetadata {
 
         public double getLatitudeAsDegreesNorth() throws ImagingException {
             final double result = latitudeDegrees.doubleValue() + latitudeMinutes.doubleValue() / 60.0 + latitudeSeconds.doubleValue() / 3600.0;
-
-            if (latitudeRef.trim().equalsIgnoreCase("n")) {
+            /* Due to some HEIC file converters to JPG longitude and latitude strings gets converted from UNICODE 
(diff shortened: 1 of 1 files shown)


garydgregory:
This doesn't look right to me and will break existing applications that have leading or trailing whitespace in that data.

Are you saying that some other app is creating bad data?

clagenna:
Well, if it'so we can insert a trim() just before substring(). Forse the
sake of completness I've used paintshop-pro to convert heic file to jpg.

Il Mer 14 Mag 2025, 18:27 Gary Gregory ***@***.***> ha
scritto:

garydgregory:
Hello @clagenna 
Let me understand: 
- you have a HEIC file
- you use PaintShopPro to convert it to a JPG
- you use Commons Imaging to read the JPG

Does the data start bad in the HEIC file?

Does PSP create a bad file?

Can you provide a unit test with each file that demonstrates the issue? Ideally the test should fail if the changes to 'main' aren't applied.

Ty.

PR #552, opened 2026-06-05
Fix int overflow in PCX image buffer sizing, now throws AllocationRequestException

alhudz:
Reading the PCX decoder I noticed `getBufferedImage` sizes the pixel buffer with int multiplies of `xSize`/`ySize`, which both come from uint16 header fields and can reach 65536. For 24/32-bit and 8-bit/3-plane images `rowLength * ySize`, `xSize * ySize` and `ySize * bytesPerImageRow` overflow int and wrap small, so the value slips past the `Allocator` 1 GB guard and an undersized buffer feeds an out-of-bounds `arraycopy`. Computing in long lets the existing `Allocator.byteArray(long)` overload reject the real size instead.

### DIFF
--- a/src/main/java/org/apache/commons/imaging/formats/pcx/PcxImageParser.java
+++ b/src/main/java/org/apache/commons/imaging/formats/pcx/PcxImageParser.java
@@ -311,7 +311,7 @@ public class PcxImageParser extends AbstractImageParser<PcxImagingParameters> {
         if ((pcxHeader.bitsPerPixel == 1 || pcxHeader.bitsPerPixel == 2 || pcxHeader.bitsPerPixel == 4 || pcxHeader.bitsPerPixel == 8)
                 && pcxHeader.nPlanes == 1) {
             final int bytesPerImageRow = (xSize * pcxHeader.bitsPerPixel + 7) / 8;
-            final byte[] image = Allocator.byteArray(ySize * bytesPerImageRow);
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
@alhudz 
There is no test for this PR.

alhudz:
Added one in `PcxImageParserTest`. It feeds a crafted 128-byte header with `bitsPerPixel=24`, `nPlanes=1` and `xMax=yMax=65535` (so `xSize=ySize=65536`), which makes `rowLength * ySize` (196608 * 65536) overflow `int` and wrap small.

- expected: `AllocationRequestException` from the 1 GB `Allocator` guard
- before the fix: the wrapped size allocates an undersized buffer and you get an `ArrayIndexOutOfBoundsException` from the `arraycopy`
- after: the size is computed in `long` so the guard sees the real ~12 GB request and rejects it

Verified the test fails on `master` and passes with the fix.

garydgregory:
@alhudz 
As stated in the PR template, run 'mvn' by itself, and fix issues _before_ you push.

alhudz:
Good catch. Running `mvn` flagged a checkstyle `LineLength` violation in the test javadoc (line was 161 chars, limit 160). I've wrapped it and the default goal is green now: `clean verify apache-rat:check checkstyle:check spotbugs:check pmd:check` all pass with the new test included.

